PR #23, opened 2021-11-28
fix govern and swap page

dansofter:
- enabled govern and swap page
- convert pools page to swap page


Kevingislason (commented):
Some of the behavior on the swap page is buggy. If you click the "to" asset dropdown, the options appear under the "from" dropdown. Also, if you click the up-down arrow icon, both assets get 'deselected.'

It would also be nice to display 'real' asset balances, as we do on the Transact page.

I am going to merge this so that we have it in place for the demo tomorrow. @dansofter Maybe you could handle these comments in a followup PR?

PR #28, opened 2021-11-30
fix_to_private_res_handle

Kevingislason:
This was causing a bug where the loader on the "To private" tab would spin indefinitely even after the transaction succeeded. The removed code was meant to handle transaction batches containing more than one transaction, but that will no longer ever happen on the "To private" tab.


dansofter (approved):
lgtm

PR #31, opened 2021-11-30
recase context files

Kevingislason:
Local changes to file names on Mac don't get committed by default, which was causing annoying bug


dansofter (approved):
lgtm

PR #33, opened 2021-12-01
mantain selected asset type/tab idx on toggle priv

Kevingislason:
Currently, when user clicks the public-private toggle, asset type and selected tab (e.g. "Send", "Receive") get set back to defaults (Polkadot and Send tab). This changes that behavior to persist tab state and asset type when the toggle is clicked.


src/App.js
```diff
@@ -7,13 +7,30 @@ import { ExternalAccountContextProvider } from 'contexts/externalAccountContext'
 import DeveloperConsole from 'components/elements/Developer/DeveloperConsole';
 import config from 'config';
 import { TxStatusContextProvider } from 'contexts/txStatusContext';
+<<<<<<< HEAD
```
dansofter, line 10: @Kevingislason can you remove this?

dansofter (changes requested):

dansofter (approved):
lgtm

PR #35, opened 2021-12-01
Fix 404 on refresh page (on gh-pages)

Kevingislason:
This fixes a bug affecting the front-end deploy on gh-pages. gh-pages expects static html i.e. no url parameters. Using react router's HashRouter fixes this because things after the # don't get sent to the server (e.g. https://dapp-alpha.manta.network/#/transact)

Also adds gh-pages and deploy command


dansofter (approved):
lgtm

PR #37, opened 2021-12-02
fix fail to conntect to signer

Kevingislason:
Currently the deployed front-end can't connect to a local instance of manta-signer. Basically, the browser if removing request origin header because it doesn't trust the signer because it is on http (being localhost) instead of https.

Allowing connection also depends on a small change to manta-signer (add the deployed front-end to allowed origins).

Small related issue: Brave blocks connections to the signer by default, I guess it thinks its an ad tracker or something. You can "lower" your "shields" if you use brave to get around this.


dansofter:
> Currently the deployed front-end can't connect to a local instance of manta-signer. Basically, the browser if removing request origin header because it doesn't trust the signer because it is on http (being localhost) instead of https.
> 
> Allowing connection also depends on a small change to manta-signer (add the deployed front-end to allowed origins).
> 
> Small related issue: Brave blocks connections to the signer by default, I guess it thinks its an ad tracker or something. You can "lower" your "shields" if you use brave to get around this.

@Kevingislason Does this fix that issue on other browsers like chrome and firefox?

dansofter (approved):
lgtm

PR #39, opened 2021-12-03
disable sidebar links & add disabled styles

Kevingislason:
-prevent user from changing page while a transaction is processing
-add disabled styles for all disabled components (except for the account select dropdown, which I couldn't figure out)
-reduce duplication of UI disabling logic


dansofter (approved):
lgtm

PR #42, opened 2021-12-04
CI pipeline

Kevingislason:
Add a basic CI pipeline that lints PRs to master branch (and also fixes all current lint errors). We can use this later as a base to do more interesting things.


dansofter (approved):
lgtm

PR #52, opened 2021-12-08
Fix/polkadotjs connection

Kevingislason:
-Fix bug with polkadot.js intermittently not connecting
-Fix bug with page crashing on changing asset types
-Fix prop type errors to declutter console while debugging the above


dansofter (approved):
lgtm

PR #56, opened 2021-12-10
Info modal if user missing required software

Kevingislason:
-User gets an info modal with download links if they are missing Manta Signer or polkadot.js wallet
-Fix bug causing occasional and unpredictable crashing (due to keyring getting initialized twice)


dansofter (approved):
lgtm

Issue #55, opened 2021-12-09
Subscan integration


stechu:
show the transaction in https://dolphin.subscan.io/ after successful sending.

PR #57, opened 2021-12-13
add subscan url to transaction success toast

dansofter:
- Add subscan url to transaction success toast
- Create NotificationContent component

Closing #55

### DIFF
--- a/src/utils/api/MakeTxResHandler.js
+++ b/src/utils/api/MakeTxResHandler.js
@@ -12,8 +12,7 @@ export function makeTxResHandler(
   return async ({ status, events }) => {
     let error,
       extrinsic = '';
-    if (status.isInBlock) {
-    }
+
     if (status.isInBlock || status.isFinalized) {
       error = getFailedExtrinsicError(events, api);
     }
--- a/src/utils/api/MakeTxResHandler.js
+++ b/src/utils/api/MakeTxResHandler.js
@@ -4,14 +4,16 @@ import getFailedExtrinsicError from './GetFailedExtrinsicError';
 export function makeTxResHandler(
   api,
   /* eslint-disable no-unused-vars */
-  onSuccess = (block) => null,
+  onSuccess = (block, extrinsic) => null,
   onFailure = (block, error) => null,
   onUpdate = (message) => null
   /* eslint-enable no-unused-vars */
-
 ) {
-  return ({ status, events }) => {
-    let error;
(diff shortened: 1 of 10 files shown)


Kevingislason (commented):
One thought--I think we probably want to link to the extrinsic rather than the block that contains it. Eventually, when we have a bunch of transactions in the same block, and it will be impossible for the user to figure out which transaction they sent. 

I think you can get the extrinsic hash when the transaction is built, before it is sent. https://polkadot.js.org/docs/api/cookbook/tx

You would probably then save the extrinsic hash with `useRef`, and then access that inside the tx success handler function.

src/utils/api/MakeTxResHandler.js
```diff
@@ -20,7 +21,18 @@ export function makeTxResHandler(
...
+        const signedBlock = await api.rpc.chain.getBlock(status.asFinalized);
+        const extrinsics = signedBlock.block.extrinsics;
+
+        extrinsic =
+          extrinsics && extrinsics.length > 0
+            ? extrinsics[extrinsics.length - 1].hash.toHex()
+            : '';
+      } catch (err) {
+        console.err(err);
+      }
```
Kevingislason, line 34: I don't think this method of getting the extrinsic hash will work. There is no guarantee that last extrinsic in a block will be the one that the user sent. 

You can probably look at the extrinsics more closely to figure out which one belongs to the user, but it's easier to get the extrinsic hash when we send the transaction. Partial example:

```js
 const ExampleSendComponent = ( ) => {
    import { useRef } from 'React';
    const extrinsicHash = useRef(null);
   
   const onTxSuccess = () => {
      showSuccess(extrinsicHash.current)
  }

   const onClickSubmit {
       // build transaction, do other things
      // [...]
       extrinsicHash.current = transaction.signAndSend(externalAccountSigner, txResHandler)
    }

}
```

dansofter: @Kevingislason Do we have a running example of this?
I don't think signAndSend will return the hash if we pass the statusCallback as the second parameter. It just returns a function that can be used to unsubscribe from the transaction status updates.

I was able to get the hash by passing {nonce: -1} option to signAndSend instead of txResHandler, but as we are not passing the statusCallback, the transaction events are not being handled correctly.

```
const txHash = await tx.signAndSend(externalAccountSigner, { nonce: -1 })
console.log({ txHash: txHash.toHuman() });
```

PR #58, opened 2021-12-13
fix overlapping refreshes

Kevingislason:
A user's private balance should refresh on every new block, *unless* there is already a refresh in progress. This PR prevents concurrent refreshes. This was causing an issue with manta-signer where a backlog of refreshes would develop.


dansofter (approved):
lgtm

PR #59, opened 2021-12-15
save last used wallet

Kevingislason:
Default to the user's most recently used polkadot.js account (as opposed to the user's first account)


dansofter (approved):
LGTM!

PR #63, opened 2021-12-20
add public accounts label

Kevingislason:
Label the public accounts dropdown so that users confuse public and private accounts somewhat less


dansofter (approved):
LGTM

PR #64, opened 2021-12-22
Feature/incremental recovery

Kevingislason:
-Refresh wallet balances incrementally to improve speed / resource use (See https://github.com/Manta-Network/manta-js/pull/20)
-Check that the user has the minimum required version of `manta-signer`
-Small change to `UpdateRequiredSoftwareModal`, so that it closes on its own when desirable
-Make notifications go away after 15 seconds

todo: update `manta-signer` npm package to latest version after publishing it (this is why lint is failing)


src/types/Version.js
```diff
@@ -0,0 +1,30 @@
...
+      this.patchVersion >= other.patchVersion
+    ) {
+      return true;
+    }
+    return false;
+  }
+
+  toString() {
+    return `v${this.majorVersion}.${this.minorVersion}.${this.patchVersion}`;
+  }
```
Kevingislason, line 29: @dansofter I removed the validity check you had here, I think we can assume that the version in the config will always be well formed. Let me know if I am missing something.

dansofter:
Looking good.
Once we fix the signer-interface package in package.json, we can merge this PR.

@Kevingislason

PR #68, opened 2022-01-12
feature copy receive address to clipboard

dansofter:


Kevingislason (changes requested):
Great stuff. Two small suggestions:

1. It would be nice to add some kind of visual cue that the address is copy-pasteable. I think the easiest way to do this is to style the address text like a hyperlink, so underlined + an accept color (I guess probably the teal color we use for the sidebar icons when selected)

2. I think the timeout is a little long, I think two seconds would be okay.

Kevingislason (commented):
New styles look good. You can remove ` hover:` imo, we want this copy-paste feature to be as obvious as possible.

Also, sorry, should have mentioned earlier, but can I suggest some spacing tweaks?

1. Probably want to horizontally center the "Address copied to clipboard" text (instead of using margin). If you shrink your browser window, it messes with the spacing of elements and the text goes off center.

2. When I click on the address link text, the bottom of the "Transact" component and the wallet icon move up, since there are fewer rows of text. The UX will look a little more polished if we can prevent this behavior.

Kevingislason (approved):
Looks great! Would you mind bumping the "Address Copied to Clipboard" text slightly to the left and making the text lowercase ("Address copied to clipboard")? After that, LGTM as far as I am concerned.

PR #73, opened 2022-01-21
custom node selector

dansofter:
- node selector component
- add api disconnected state


src/components/elements/Layouts/Navbar/NodeSelector.js
```diff
@@ -0,0 +1,90 @@
+import React, { useEffect, useState } from 'react';
+import OutsideClickHandler from 'react-outside-click-handler';
+
+import { useSubstrate } from 'contexts/substrateContext';
+import { dolphinNodes } from 'constants/NodeConstants';
```
Kevingislason, line 5: Probably best to avoid variables with "dolphin" in the name, because in the long term we want this same repo to work for all of Dolphin, Calamari, and Manta.

src/components/elements/Layouts/Navbar/NodeSelector.js
```diff
@@ -0,0 +1,90 @@
...
+    setNodeSelected(node);
+  };
+
+  return (
+    <div className="mr-auto relative">
+      <p
+        className="text-primary text-sm pb-1 cursor-pointer"
+        onClick={() => setShowPopup(true)}
+      >
+        Dolphin Node <FontAwesomeIcon icon={faChevronCircleDown} />
```
Kevingislason, line 26: Similar to above comment, we probably want to avoid hard-coding the word "Dolphin." You can handle this through configs instead, so that it is easier to add Manta and Calamari configs in the future:
1. `import config from 'config'`
2. add `NETWORK_NAME: "Dolphin"` to `src/config/production.json` and `NETWORK_NAME: "Local testnet"` to `src/config/development.json`
3. On this line, replace "Dolphin Node" with `${config.NETWORK_NAME} Node`

src/constants/NodeConstants.js
```diff
@@ -0,0 +1,34 @@
+export const dolphinNodes = [
```
Kevingislason, line 1: Similar to above comment, this lists of nodes should be determined by a config file. If I am using the `production.json` config file, I should see anjie, eddie, etc.  If I am using the `development.json` config file, I should see a list containing only the default local node:
```
name: 'local-default'
url: 'ws://localhost:9944'
```

Kevingislason (changes requested):
Very cool

Some UI level suggestions:
-Brandon had earlier suggested we use checkmarks and Xs instead of red and green circles for signer connection status--if you are colorblind, these colored circles are difficult to interpret. So I think we should probably stick to that convention here, though I quite like the flashing light effect.

-On light mode, the dropdown has a teal background, which I think is too dark / too much of an accent color, we probably want this to be some shade of gray. But the current dark mode color looks nice, lets keep that.

-When I switch nodes, the UI shows disconnection temporarily. I think it would be a little more slick to display a tiny loader for ~1 second when the user changes nodes, so that it doesn't look like anything is wrong.

-I think it would look very slick if the chevron pointed right when the node selection popup is closed, and down when the popup is open.


Maybe you recall that we also want the user to be able to add a custom node url (a local node for example). All good if you want to do that in a separate PR.

Btw, you have a bunch of lint errors. I think your local eslint config is out of sync with the repo.

Kevingislason (approved):
LGTM

PR #75, opened 2022-01-26
add custom node input

dansofter:
- add custom node input
- disconnect previous api instance before connecting to new endpoint


src/contexts/substrateContext.js
```diff
@@ -69,7 +74,9 @@ const connect = (state, dispatch) => {
   });
   _api.on('ready', () => dispatch({ type: 'CONNECT_SUCCESS' }));
   _api.on('error', (err) => dispatch({ type: 'CONNECT_ERROR', payload: err }));
-  _api.on('disconnected', () => dispatch({ type: 'DISCONNECTED' }));
+  _api.on('disconnected', () => {
+    if (apiState !== 'READY') dispatch({ type: 'DISCONNECTED' });
```
Kevingislason, line 78: This is to prevent the flickering X on change node, right? The problem with this approach is that
`if (apiState !== 'READY')` means that the UI won't show if the api was connected and then gets disconnected, for example if a node goes down, or the user loses their wifi connection.

I think you can instead check if the current `socket` is the same as `api`'s url. Alternatively, you could handle this with a timer inside `NodeSelector.js`--if node is disconnected, show the loader and set a timeout. At the end of the timeout, the timeout's callback checks if the api is still disconnected, and only then shows the red X.

src/components/elements/Layouts/Navbar/NodeSelector.js
```diff
@@ -1,42 +1,58 @@
...
-      <p
+    <div className="relative -ml-3 border border-primary  dark:border-white py-1 px-2 rounded-lg">
+      {/* <p
         className="text-primary text-sm pb-1 cursor-pointer"
         onClick={() => setShowPopup(true)}
       >
         {config.NETWORK_NAME} Node{' '}
         <FontAwesomeIcon
           icon={showPopup ? faChevronCircleDown : faChevronCircleRight}
         />
```
Kevingislason, line 52: can delete this now I think

Kevingislason (changes requested):

Kevingislason (approved):

PR #88, opened 2022-05-04
add extrinsic to successful transaction notification

dansofter:


Kevingislason (approved):
I made one tweak: the way this was implemented this grabbed the last extrinsic in the block and assumed it belonged to the user, which isn't correct if there are several extrinsics in the block. Otherwise, LGTM

Issue #85, opened 2022-04-30
Inconsistent spacing on `Send Max Amount` button


bhgomes:
The space between `Send` and `Max` is smaller than the space between `Max` and `Ammount`

<img width="958" alt="Screen Shot 2022-04-30 at 12 52 57" src="https://user-images.githubusercontent.com/9126085/166114935-569fcef0-1261-41e4-abd3-8977cab658cb.png">

PR #102, opened 2022-05-18
Testnet v2 receiver public balance

dansofter:
- add receiver public balance component
- fix Max Amount spacing #85


Kevingislason (approved):
Looks good. Added one small tweak so that the '...' loader doesn't show up if there is no address.

PR #118, opened 2022-05-20
Front end network-specific configuration

dp3000:
- [X] Add new configuration files for **dolphin**, **calamari**, and **manta** networks
- [X] Modify `src/config/index.ts` to load the default network configuration at runtime
- [X] Refactor hardcoded network references out of `public/index.html` and `src/components/Sidebar/index.tsx`
- [X] Add new `Logo` component that renders appropriate logo given the network configuration
- [X] Setup network-specific `.scss` files where any network-specific styling can go
- [X] Rename and add `.png`s for each of the different network logos


bhgomes:
Thanks for the config file setup! Can we also add at least the following to the configurations?
- blockchain node URLs
- color scheme
Once we have some more details on the network-specific differences we can do more rounds of PRs.

src/constants/NetworkConstants.ts
```diff
@@ -0,0 +1,6 @@
+// @ts-nocheck
+export const network = {
+  Name: 'Dolphin',
+  Title: 'Dolphin Testnet App',
+  Url: 'https://dolphin.subscan.io/block',
+};
```
bhgomes, line 6: Thanks for the config file setup! Can we also add at least the following to the configurations?
- blockchain node URLs
- color scheme

Once we have some more details on the network-specific differences we can do more rounds of PRs.

src/constants/NetworkConstants.ts
```diff
@@ -0,0 +1,6 @@
+// @ts-nocheck
+export const network = {
+  Name: 'Dolphin',
```
bhgomes, line 3: Can we use `camelCase` variable names?

bhgomes (changes requested):
This is great, thanks Drew! Can you add a switch on the frontend to choose which configuration to use at runtime? Also, please fill out a PR description so we can understand the changes at a high level.

src/constants/NetworkConstants.ts
bhgomes: Dolphin, Calamari, and Manta may all be using different styling like foreground/background colors, icons, etc.

src/config/calamari.json
```diff
@@ -0,0 +1,36 @@
...
+    {
+      "name": "roosta",
+      "url": "wss://roosta.rococo.calamari.engineering "
+    },
+    {
+      "name": "zaphod",
+      "url": "wss://zaphod.rococo.calamari.engineering "
+    }
+  ],
+  "PAGE_TITLE": "Calamari Testnet App",
```
bhgomes, line 33: This should just be Calamari as this is a deployment not a testnet

bhgomes (changes requested):

tommyjk21 (approved):

tommyjk21 (approved):

src/config/common.json
```diff
@@ -1,6 +1,6 @@
 {
-    "APP_NAME": "manta-web-app",
-    "RPC": {},
-    "SIGNER_URL": "http://localhost:29987/",
-    "MIN_REQUIRED_SIGNER_VERSION": "0.6.0"
+  "APP_NAME": "manta-web-app",
+  "MIN_REQUIRED_SIGNER_VERSION": "0.6.0",
+  "RPC": {},
```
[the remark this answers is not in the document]
bhgomes: If it isn't used for anything then we can get rid of it.

bhgomes (approved):

PR #127, opened 2022-05-27
use load balancer

Kevingislason:
Use a load balancer instead of a hardcoded list of nodes defaulting to `Ford`

### DIFF
--- a/src/config/development.json
+++ b/src/config/development.json
@@ -1,37 +1,12 @@
 {
-    "PROVIDER_SOCKET": "wss://ford.rococo.dolphin.engineering",
+    "PROVIDER_SOCKET": "wss://ws.rococo.dolphin.engineering",
     "DEV_CONSOLE": true,
     "DEVELOPMENT_KEYRING": false,
     "BASE_STORAGE_KEY": "dolphinDev",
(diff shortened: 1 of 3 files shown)


tommyjk21 (approved):

src/config/development.json
```diff
@@ -1,37 +1,12 @@
 {
-    "PROVIDER_SOCKET": "wss://ford.rococo.dolphin.engineering",
+    "PROVIDER_SOCKET": "wss://ws.rococo.dolphin.engineering",
     "DEV_CONSOLE": true,
     "DEVELOPMENT_KEYRING": false,
```
bhgomes, line 4: why is this `false`?

bhgomes (changes requested):

src/config/development.json
Kevingislason: DEVELOPMENT_KEYRING is default accounts like Alice and Bob, we can get rid of this

stechu (approved):

src/config/development.json
Kevingislason: Err, actually we might use this in a selenium testing config

bhgomes (approved):

PR #145, opened 2022-06-09
add initial sync state, add animation to ...

dansofter:
## Description

- Added initial sync state to the private wallet context
- Added animation to ... balance string

---

- [ ] Test following procedure in docs/manual_test.md
- [ ] Updated relevant documentation in the code.
- [ ] Re-reviewed Files changed in the Github PR explorer.
- [ ] Update the version number in package.json


src/pages/SendPage/SendToForm/SendToAddressForm.tsx
```diff
@@ -103,10 +104,19 @@ const SendToAddressForm = ({
...
+    receiverAddress,
+    isToPrivate,
+    isPrivateTransfer
+  } = useSend();
+  const { isInitialSync, walletIsBusy } = usePrivateWallet();
 
   let balanceString;
-  if (receiverAddress && receiverCurrentBalance) {
+  if (isInitialSync && (isToPrivate() || isPrivateTransfer())) {
+    balanceString = 'Syncing With Ledger';
```
Kevingislason, line 118: could we change this to text slightly to `Syncing to ledger...` ?

src/pages/SendPage/SendToForm/SendToAddressForm.tsx
```diff
@@ -117,7 +127,16 @@ const ReceiverBalanceDisplay = () => {
   return (
     <div className="flex justify-between items-center px-6 py-2">
       <div className="text-black dark:text-white">
-        Balance: <strong>{balanceString}</strong>
+        Balance:{' '}
+        {walletIsBusy.current && !isInitialSync ? (
```
Kevingislason, line 131: Two things:
-It might be good to factor the loader out into its own react component--this can avoid the duplication that codeclimate mentions
-we don't want the '...' loader to show whenever the wallet is syncing, only when the wallet is syncing AND a current balance is not available

Kevingislason (changes requested):

src/components/Balance/index.tsx
```diff
@@ -0,0 +1,39 @@
...
+
+const Balance: React.FC<IBalanceProps> = ({
+  balance,
+  className,
+  loaderColor
+}) => {
+  return (
+    <div className={className}>
+      Balance:&nbsp;
+      {balance === '...' ? (
```
Kevingislason, line 17: When we want to show the loader, I think it makes more sense to pass in a `null` balance rather than the string `...`

src/pages/SendPage/SendToForm/SendToAddressForm.tsx
```diff
@@ -103,22 +106,31 @@ const SendToAddressForm = ({
...
+  const { isInitialSync } = usePrivateWallet();
 
   let balanceString;
-  if (receiverAddress && receiverCurrentBalance) {
-    balanceString = receiverCurrentBalance.toString();
+  if (isInitialSync && (isToPrivate() || isPrivateTransfer())) {
+    balanceString = 'Syncing to ledger...';
   } else if (receiverAddress) {
-    balanceString = '...';
+    balanceString = getBalanceString(receiverCurrentBalance);
```
Kevingislason, line 122: I don't think the `getBalanceString` function is necessary anymore, I think you can just do `balance?.toString()` and pass `null` into the balance component as I suggest in the first comment

src/pages/SendPage/SendFromForm/SendAmountInput.tsx
```diff
@@ -9,15 +9,23 @@ import Balance from 'types/Balance';
...
-    getMaxSendableBalance
+    getMaxSendableBalance,
+    isToPublic,
+    isPrivateTransfer
   } = useSend();
-  const balanceText = getBalanceString(senderAssetCurrentBalance);
+  const { isInitialSync } = usePrivateWallet();
+  const balanceText =
+    isInitialSync && (isPrivateTransfer() || isToPublic())
+      ? 'Syncing to ledger...'
```
Kevingislason, line 27: Would it be possible to add the animation to this `...`?

Kevingislason (changes requested):
Looking nice! Just suggesting one small refactor here. Would also be cool if we could add the `...` animation to  `Syncing to ledger...`

src/components/Balance/index.tsx
```diff
@@ -0,0 +1,39 @@
...
+  loader
+}) => {
+  return (
+    <div className={className}>
+      Balance:&nbsp;<strong>{balance}</strong>
+      {loader ? (
+        <div className="inline-flex items-center">
+          <div
+            className="dot1"
+            style={{ backgroundColor: loaderColor, borderColor: loaderColor }}
```
Kevingislason, line 23: It's preferable to do this styling using tailwind classNames instead of these inline styles if possible

Kevingislason (changes requested):
I noticed that that the receiver balance loader doesn't show up on light mode. Otherwise, just a few code cleanup things.

Kevingislason (approved):

bhgomes (approved):

Issue #205, opened 2022-07-08
Handle sync errors, surface to the UI


Kevingislason:
Currently, when synchronization fails due networking issues or other exceptions, the error is not surfaced to the user and the loader spins on indefinitely. Users might waste hours waiting for a sync to finish when in reality the sync has totally errored out. Errors should be handled gracefully and surfaced through the UI

Issue #172, opened 2022-06-16
Progress bar for sync


Kevingislason:
When syncing to chain, users should get a loading bar that let's them know roughly what percent of the sync is finished

PR #209, opened 2022-07-20
add sync status

dansofter:
## Description

closes: #205 #172

---

## If PR to `staging`


src/contexts/privateWalletContext.tsx
```diff
@@ -39,13 +44,120 @@ export const PrivateWalletContextProvider = (props) => {
...
+  const initialSyncReceivers = useRef(0);
+  const currentSyncReceivers = useRef(0);
+  const syncSenderIndex = useRef(0);
+  const syncPercentage = useRef(0);
+  const nextSyncPercentage = useRef(0);
+  const pullBatchStartTime = useRef(0);
+  const pullBatchEndTime = useRef(0);
+  const percentagePerBatch = useRef(0);
+  const timePerPercentage = useRef(0);
+  const syncError = useRef(false);
```
Kevingislason, line 60: Also, do all of these really need `useRef`?

Kevingislason, line 60: I think it makes more sense to use null as a default rather than 0 for these, it's less ambiguous / confusing in the initial case where you just don't know what these values are.

Kevingislason, line 60: Are you sure all of this state is necessary? Some state such as `syncPercentage`, `nextSyncPercentage`, `timePerPercentage`, and `percentagePerBatch` can be computed purely from other state. You can define functions to get these values and just store them in normal constants:
```js
const percentagePerBatch = getPercentagePerBatch( );
```

src/contexts/privateWalletContext.tsx
```diff
@@ -39,13 +44,120 @@ export const PrivateWalletContextProvider = (props) => {
   const [signerVersion, setSignerVersion] = useState(null);
   const [isReady, setIsReady] = useState(false);
   const [isInitialSync, setIsInitialSync] = useState(false);
+  const [error, setError] = useState(false);
```
Kevingislason, line 47: I think `initWalletError` is more specific

tailwind.config.js
```diff
@@ -50,7 +50,9 @@ module.exports = {
         current: 'var(--color-text-secondary)',
         primary: 'var(--color-fill-primary)',
         secondary: 'var(--color-fill-secondary)',
-        gray: 'var(--color-fill-gray)'
+        gray: 'var(--color-fill-gray)',
+        black: '#000',
+        white: '#fff'
```
Kevingislason, line 55: I don't think these are used anywhere, and tailwind already has `text-black` (or white) and `bg-black` (or white)

Kevingislason (changes requested):

src/constants/LocalStorageConstants.ts
```diff
@@ -4,4 +4,6 @@ export const localStorageKeys = {
   CurrentToken: 'current-token',
   IsPrivateReceiver: 'is-private-receiver',
   IsPrivateSender: 'is-private-sender',
+  CurrentSyncReceiversCount: 'current-sync-receivers-count',
+  CurrentSyncSenderIndex: 'current-sync-sender-index'
```
bhgomes, line 8: Why do receivers and senders get a different name for these constants?

src/contexts/privateWalletContext.tsx
```diff
@@ -39,13 +44,120 @@ export const PrivateWalletContextProvider = (props) => {
...
+            : parseInt(syncPercentage.current) +
+              parseInt(percentagePerBatch.current);
+      } else {
+        syncPercentage.current = 0;
+      }
+      pullBatchStartTime.current = performance.now();
+      syncSenderIndex.current = sender_index;
+    } catch (err) {
+      console.log('sync callback error - ', err);
+    }
```
bhgomes, line 131: Is this the standard formatting in JS? seems all over the place. Which formatter should we use @Kevingislason ?

Kevingislason: We should use the eslint config defined in `.eslintrc` in the root directory

package.json
```diff
@@ -1,86 +1,86 @@
...
+    "bn.js": "^5.2.0",
+    "classnames": "^2.3.1",
+    "coin-selection": "^0.1.1",
+    "decimal.js": "^10.3.1",
+    "element-react": "^1.4.34",
+    "element-theme-default": "^1.4.13",
+    "gh-pages": "^3.2.3",
+    "husky": "^7.0.1",
+    "manta-wasm-wallet": "1.4.0",
+    "manta-wasm-wallet-api": "1.5.1",
```
Kevingislason, line 34: You should use the same version for both of these libraries

src/components/DotLoader/index.tsx
```diff
@@ -0,0 +1,11 @@
...
+
+const DotLoader = () => {
+    return <div className="inline-flex items-center">
+        <div className={`loader-dot1 bg-black dark:bg-white`} />
+        <div className={`loader-dot2 bg-black dark:bg-white`} />
+        <div className={`loader-dot3 bg-black dark:bg-white`} />
+    </div>
+}
+
+export default DotLoader;
```
Kevingislason, line 11: I think this should go in a more general directory inside of `components` called `Loaders`

src/components/SyncPercentage/index.tsx
```diff
@@ -0,0 +1,26 @@
+import React from 'react';
```
Kevingislason, line 1: As above, I think this should go in a more general directory inside of `components` called Loaders

src/contexts/privateWalletContext.tsx
```diff
@@ -74,6 +186,7 @@ export const PrivateWalletContextProvider = (props) => {
     };
 
     const initWallet = async () => {
+      try {
```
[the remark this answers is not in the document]
Kevingislason: Please get rid of this try catch or tell me why I am wrong

src/contexts/privateWalletContext.tsx
```diff
@@ -74,32 +87,54 @@ export const PrivateWalletContextProvider = (props) => {
     };
 
     const initWallet = async () => {
```
Kevingislason, line 89: This function is getting insanely long, I think it's time to factor out some subfunctions; see comments below for some suggestions

src/contexts/privateWalletContext.tsx
```diff
@@ -74,32 +87,54 @@ export const PrivateWalletContextProvider = (props) => {
...
+          api,
+          externalAccountSigner,
+          DEFAULT_PULL_SIZE,
+          DEFAULT_PULL_SIZE,
+          updateSyncProgress,
+          syncErrorCallback
+        );
+        const wasmApi = new Api(wasmApiConfig);
+        const wasmLedger = new wasm.PolkadotJsLedger(wasmApi);
+        const wasmWallet = new wasm.Wallet(wasmLedger, wasmSigner);
```
Kevingislason, line 108: I think this could be a subfunction called `createWasmWallet`

src/contexts/privateWalletContext.tsx
```diff
@@ -74,32 +87,54 @@ export const PrivateWalletContextProvider = (props) => {
...
+
+        pullBatchStartTime.current = performance.now();
+        currentSyncReceivers.current = store.get(
+          localStorageKeys.CurrentSyncReceiversCount,
+          0
+        );
+        syncSenderIndex.current = store.get(
+          localStorageKeys.CurrentSyncSenderIndex,
+          0
+        );
```
Kevingislason, line 118: I think this could be a subfunction (expoerted from `privateWalletSyncContext`) called `initSyncState`

src/contexts/privateWalletContext.tsx
```diff
@@ -74,32 +87,54 @@ export const PrivateWalletContextProvider = (props) => {
...
+
+        const privateAddress = await getPrivateAddress(wasm, wasmWallet);
+        setPrivateAddress(privateAddress);
+        console.log('Beginning initial sync');
+        const startTime = performance.now();
+        await wasmWallet.restart();
+        const endTime = performance.now();
+        console.log(
+          `Initial sync finished in ${(endTime - startTime) / 1000} seconds`
+        );
```
Kevingislason, line 128: I think this could be a subfunction called `restart`

src/contexts/privateWalletSyncContext.tsx
```diff
@@ -0,0 +1,144 @@
...
+
+export const PrivateWalletSyncContextProvider = (props) => {
+  // sync state
+  const currentSyncReceivers = useRef(0);
+  const syncSenderIndex = useRef(0);
+  const pullBatchStartTime = useRef(0);
+  const pullBatchEndTime = useRef(0);
+  const [syncPercentage, setSyncPercentage] = useState(0);
+  const [nextSyncPercentage, setNextSyncPercentage] = useState(0);
+  const [timePerPercentage, setTimePerPercentage] = useState(0);
```
Kevingislason, line 23: All of these should be initialized to `null` instead of `0`, since they are unknown values

src/contexts/privateWalletSyncContext.tsx
```diff
@@ -0,0 +1,144 @@
...
+        );
+      } else {
+        setSyncPercentage(0);
+      }
+      pullBatchStartTime.current = performance.now();
+      syncSenderIndex.current = sender_index;
+    } catch (err) {
+      console.log('sync callback error - ', err);
+    }
+  };
```
Kevingislason, line 96: This function is very long and complex, you should decompose it to separate subfunctions as much as possible, and add explanatory comments

Kevingislason (changes requested):
This is coming along, mostly just need to refactor some of the long / complex functions.

You also have some merge conflicts to fix btw.

Also please respond to Brandon's comments.

src/contexts/privateWalletSyncContext.tsx
Kevingislason: Start by commenting the function, then I can make suggestions.

src/contexts/privateWalletSyncContext.tsx
Kevingislason: This is a convention we use throughout the front end, and a best practice for React state in general. If a value comes from an API call, null is generally used as a default. 

UI behavior is not the point. Strictly speaking, initial values of zero not correct. For example `currentSyncReceivers` does not "really" equal zero on pageload, because there is not a current sync immediately on pageload. State should always be accurate, even when inaccuracy doesn't cause UI problems. Otherwise, application code is harder to read and debug.

>we will also need unnecessary null conditional statements
This is possible to avoid if you can modularize `updateSyncProgress`. We should only have to do a single null check.

AlexParshh (dismissed):

PR #271, opened 2022-10-12
fix staking ui feedback

dansofter:
## Description

-Fixes issue with urls ignoring params before '#' by getting rid of hash router
-Show only active collators by default
-Reduce staking component width to improve UX on narrower screens
-Adds balance display to stake modals
-Adds call to action for users to run collators
-Adds APY display to staking table

---

## If PR to `staging`

## If PR to `main`
- [ ] Update `CHANGELOG.md`


src/pages/StakePage/Tables/StakingTable.tsx
```diff
@@ -80,19 +80,33 @@ const StakingTable = () => {
...
+
   const rowData = userDelegations.map((delegation: Delegation) => {
     return {
       Collator: delegation.collator,
       Amount: delegation.delegatedBalance,
       Status: getIsEarningString(delegation),
       Rank: delegation.rank,
+      'APY Estimate': collatorCandidates.find(
+        (collator) => collator.address === delegation.collator.address
+      ),
```
Kevingislason, line 99: You also need to handle the case where the collator that the user delegated to is not in the collator candidates, i.e. the collator became inactive. In that case, you can just display 0%

Kevingislason (changes requested):

AlexParshh (approved):

src/MainApp.tsx
```diff
@@ -31,7 +31,10 @@ function MainApp() {
           />
           <Route
             path="/calamari"
-            render={() => <Redirect to="/calamari/stake" />}
+            render={(props) => {
+              console.log({ props });
```
Kevingislason: delete

Kevingislason (commented):
Mostly looks good. The one thing I'd change is that the balance display in the modal looks strange, being out of alignment with "Staked" and "Minimum Stake", I think it should go right under those.

Also wanted to tack on a few small things to this PR:
1. Remove the "Status" column from the collators table (but not the staking table); users can tell a collator is inactive from the filters, and from the APY.
2. 1^^ will allow us to make the collators table less wide, so we can reduce its width until the table cell holding the buttons is just about touching the side of the table. We can also reduce the width of the other two tables by the same amount.
3. Chloe suggested we restyle "Launch your own collator" to look like this:
 
unknown-2

Boyuan-Deng:
Nice work Danny, most of the changes look good.

Want to open a small design discussion here, in the dark mode, I would prefer a whiter color for "Active / All / Inactive" to make it stands out and match the color when there is text inside the search bar.
image
image

Boyuan-Deng (approved):
Opened a small design discussion but the rest Looks good to me

Kevingislason (approved):

src/config/development.json
```diff
@@ -2,11 +2,11 @@
     "DEV_CONSOLE": true,
     "DEVELOPMENT_KEYRING": false,
     "DOLPHIN_PROVIDER_SOCKET": "wss://ws.rococo.dolphin.engineering",
```
Boyuan-Deng, line 4: I think the dolphin_provider_socket also changed but not sure we want to include in this PR @Kevingislason [c3.dolphin.engineering](https://polkadot.js.org/apps/?rpc=wss%253A%252F%252Fc3.dolphin.engineering)

src/pages/StakePage/Tables/StakingTable.tsx
```diff
@@ -206,7 +236,7 @@ const StakingTable = () => {
 
   return (
     <>
-      <div className="mt-20">
+      <div style={{ maxWidth: MAX_TABLE_WIDTH }} className="mt-20 mx-auto">
```
Boyuan-Deng, line 239: would be best if we can avoid style and use className

src/pages/StakePage/Tables/UnstakingTable.tsx
```diff
@@ -111,7 +113,7 @@ const UnstakingTable = () => {
     return <div />;
   }
   return (
-    <div className="mt-8">
+    <div style={{ maxWidth: MAX_TABLE_WIDTH }} className="mt-8 mx-auto">
```
Boyuan-Deng, line 116: would be best if we can avoid style and use className to keep css consistent

Boyuan-Deng (approved):
Nice work Danny! Please consider the comments above. Otherwise, the code LGTM

src/config/development.json
Boyuan-Deng: Ignore this comment. I think we should include it in another PR

PR #295, opened 2022-11-08
Style/private payment redesign

Boyuan-Deng:
## Description

closes: #XXXX

---

Before we can merge this PR, please make sure that all the following items have been checked off. If any of the checklist items are not applicable, please leave them but write a little note why.

- [ ] Linked to Github issue with discussion and accepted design OR have an explanation in the PR that describes this work
- [ ] Re-reviewed files changed in the Github PR explorer

## If PR to `staging`
- [ ] Updated relevant documentation in the code
- [ ] Test code changes manually, and describe your procedure in a comment on this PR
- [ ] Mark this PR with the correct milestone

## If PR to `main`
- [ ] Update the version number in `package.json` and `src/config/common.json`
- [ ] Update the minimum required signer version number in `src/config/common.json` if applicable
- [ ] Test following the procedure in `docs/manual_test.md`
- [ ] Update `CHANGELOG.md`
- [ ] Make sure that all PR's to `manta-signer` and `sdk` on which this PR depends are merged

PR #320, opened 2022-12-01
v3  redesign

Kevingislason:
## Description

UI restyle for testnet v3

See https://github.com/Manta-Network/manta-front-end/pull/295 for full commit history

---

## If PR to `staging`


src/pages/SendPage/SendButton.tsx
```diff
@@ -53,26 +28,125 @@ const SendButton = () => {
...
+  ) {
+    validationMsg = `Invalid ${
+      isPrivateTransfer() ? 'zkAddress' : 'Substrate address'
+    }`;
+  } else if (receiverAmountIsOverExistentialBalance() === false) {
+    const existentialDeposit = new Balance(
+      receiverAssetType,
+      receiverAssetType.existentialDeposit
+    );
+    validationMsg = `min > ${existentialDeposit.toDisplayString(12, false)}`;
```
Kevingislason, line 103: I would do `Min transaction is ${existentialDeposit.toDisplayString(12, false)}` if it fits

Also, pretty sure we only need 1 or 2 or 3 decimals rather than 12

Kevingislason (commented):
- [x] I sometimes get a bug where an account displays under the wrong wallet, have you seen this? Tends to happen if I switch between talisman and subwallet right after page load. 
(BD): Do not think I have ever encountered this. Tried different ways to switch between wallets when the page loads. Would appreciate a step to replicate.

- [ ] If signer is connected but there is no public wallet connected, I still see 'Connect Signer' at the top of the screen
(BD): I looked into the codebase and it seems like privateWalletContext initWallet is dependent on externalAccountSigner so if externalAccountSigner does not exist (externalAccount not exist), signer would not be initialized. To confirm here, is it even possible to have signer when wallet not connected?

Some other minor UI things I noticed:
- [ ] our capitalization is inconsistent throughout the app. Sometimes we capitalize all words in a phrase, and sometimes we only capitalize the first word. Our design is inconsistent on this, but in general we should only capitalize the first word in a phrase, this is what everybody does.
- [x] Just realized that the size of the bridge page is a bit inconsistent with the design, in which it's about the same size as the ZkTransact component. We don't need to get it quite that thin, but we should give it a shot at least. We can remove some space around the icon and reduce the width / font size of the network select components.
- [ ] Similarly, we should make the padding of the Send and Bridge pages consistent

src/components/Accounts/AccountSelectMenu.tsx
```diff
@@ -82,7 +82,12 @@ const AccountSelectMenu = () => {
   return externalAccount ? (
     <DisplayAccountsButton />
   ) : (
-    <ConnectWallet isButtonShape={true} />
+    <ConnectWallet
+      isButtonShape={true}
+      className={
+        'bg-connect-wallet-button text-white py-3 px-4 font-medium cursor-pointer rounded-lg'
+      }
```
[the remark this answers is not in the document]
Boyuan-Deng: There are actually two. I reused this component for connect wallet button in the sendButton and that is the reason I pass in the classname instead of hard coding it.

src/pages/SendPage/SendButton.tsx
```diff
@@ -53,26 +28,125 @@ const SendButton = () => {
...
+    receiverAmountIsOverExistentialBalance,
+    senderAssetTargetBalance,
+    senderNativeTokenPublicBalance
+  } = useSend();
+  const { signerIsConnected, signerVersion } = usePrivateWallet();
+  const { externalAccount } = useExternalAccount();
+
+  let validationMsg = null;
+  let isConnectWallet = false;
+  let isConnectSigner = false;
```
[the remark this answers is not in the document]
Boyuan-Deng: I changed the naming but I still initialize them inside the if else statement because we should not check other validation condition if wallet/signer is being detected disconnected and the logic should be easier if it is inside the if else statement

src/contexts/keyringContext.tsx
```diff
@@ -25,36 +25,29 @@ export const KeyringContextProvider = (props) => {
...
-      setHasAuthToConnectSubwalletStorage(true);
-      setHasAuthToConnectSubwallet(true);
-    }
+    const walletNames = addWalletName(extensionName, hasAuthToConnectWallet);
+    setHasAuthToConnectWalletStorage(walletNames);
+    setHasAuthToConnectWallet(walletNames);
   };
 
-  const subscribeWalletAccounts = async (wallet) => {
+  const subscribeWalletAccounts = async (wallet, saveToStorage = true) => {
```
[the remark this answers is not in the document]
Boyuan-Deng: saveToStorage is sometimes set to false. When there are multiple wallets authorized to automatically connect at the page load, we want the lastAccessedWallet to be loaded at last because that is how we memorized the user's recent wallet & account selection. However, in order to prevent previously loaded wallet from resetting the parameter lastAccessedWallet, I came up with this way to prevent it. In keyring context file, line 170, you can see me set it to false.

src/pages/SendPage/SendButton.tsx
Boyuan-Deng: (reminder to myself) will need to make this changer to bridge as well

Kevingislason:
A few UI level things (will check out code tomorrow)
- [ ] The zkTransact page shows dot loader even when wallet is disconnected, instead we should just show '--'
- [ ] We should show the install public wallet modal when the user clicks the send validation button if both signer and wallet are disconnected
- [ ] Regarding "If signer is connected but there is no public wallet connected, I still see 'Connect Signer' at the top of the screen"--we don't need to display an address. Instead, we can just show an info modal that tells the user to connect their public wallet. We can re-use the same component we use for the signer out of date modal

src/pages/SendPage/SendButton.tsx
```diff
@@ -53,26 +28,141 @@ const SendButton = () => {
...
+    senderNativeTokenPublicBalance
+  } = useSend();
+  const { signerIsConnected, signerVersion } = usePrivateWallet();
+  const { externalAccount } = useExternalAccount();
+
+  let validationMsg = null;
+  let shouldShowWalletMissingValidation = false;
+  let shouldShowSignerMissingValidation = false;
+  let shouldShowWalletSignerMissingValidation = false;
+  let bgColor = 'gradient-button filter brightness-50';
```
Kevingislason, line 71: I don't think this is used

src/contexts/keyringContext.tsx
```diff
@@ -25,62 +31,63 @@ export const KeyringContextProvider = (props) => {
...
-  );
-  const [hasAuthToConnectTalisman, setHasAuthToConnectTalisman] = useState(
-    getHasAuthToConnectTalismanStorage()
-  );
-  const [hasAuthToConnectSubwallet, setHasAuthToConnectSubwallet] = useState(
-    getHasAuthToConnectSubwalletStorage()
+  const [hasAuthToConnectWallet, setHasAuthToConnectWallet] = useState(
+    getHasAuthToConnectWalletStorage()
   );
+  const keyringIsBusy = useRef(false);
```
Kevingislason, line 37: I don't really understand what this is doing, can you explain / add a comment?

src/components/Accounts/ZkAccountDisconnectedWalletModal.tsx
```diff
@@ -0,0 +1,16 @@
...
+    <div className="flex flex-col items-center w-80 mt-3 bg-primary rounded-lg p-4 absolute right-0 top-full z-50 border border-white border-opacity-20">
+      <h1 className="text-center pt-3 pb-2 text-white font-medium text-xl">
+        Connect Wallet
+      </h1>
+      <p className="py-4 pl-8 pr-5 text-xss tighter-word-space text-white text-opacity-70">
+        Connect a wallet to view your zkAddress
+      </p>
+    </div>
+  );
+};
```
Kevingislason, line 14: Since this has the same shape and style as the signer out of date modal, I think we may want to abstract out a component to use for both, and just pass in the title / text. That way, we won't need to remember to restyle multiple components if one changes.

Kevingislason:
Approved

PR #333, opened 2022-12-16
Testnet UI V3 Demo Feedback

Boyuan-Deng:
## Description

- Show abbreviated account name on the ExternalAccountBlock (top right)
- Restyle wallet dropdown to fix the background color under wallet icon
- Gray out swap sender receiver button and add the disabled cursor when it’s not interactable (i.e. public to public or private to private)
- Allow user to at least copy open the public accounts dropdown and copy public addresses when transaction is processing (but disable switching between browser wallets / addresses)


src/pages/SendPage/SendForm.tsx
```diff
@@ -52,14 +58,23 @@ const SendForm = () => {
...
-            src={Svgs.UpDownArrowIcon}
-            alt="arrow-up-down-icon"
-          />
+          <div onClick={onClickSwapSenderReceiver}>
+            <UpDownArrowSvg
+              className={classNames(
+                'mx-auto my-4 cursor-pointer',
+                {
+                  disabled: disabled,
+                  'cursor-not-allowed': disabledSwapSenderReceiver
```
Kevingislason, line 72: `disabled` and `cursor-not-allowed` are the same thing, let's just use `disabled` since it is more common throughout the app

```js
disabled: disabled || disabledSwapSenderReceiver
```

src/components/Accounts/ConnectWallet.tsx
```diff
@@ -4,22 +4,30 @@ import classNames from 'classnames';
...
+  isButtonShape,
+  setIsMetamaskSelected,
+  className = ''
+}) => {
   const { ModalWrapper, showModal } = useModal();
-  const handleOnClick = () => showModal();
+  const { txStatus } = useTxStatus();
+  const disabled = txStatus?.isProcessing();
+  const onClickPlusIcon = () => !disabled && showModal();
+
```
Kevingislason, line 19: I don't think the disabled styles here are necessary since there is no way to transact without having a wallet connected

Boyuan-Deng: `const onClickPlusIcon = () => !disabled && showModal()`
 Disabled here is trying to prevent user from connecting another wallet. 
 
The problem is that when user can connect another wallet while the transaction is processing. The newly connected wallet will try to reset the externalAccount which is not wanted.

Kevingislason: Oh sorry, makes sense

PR #349, opened 2023-01-09
warn users when privatizing fee balance

Kevingislason:
## Description

Displays a warning to users when they are about to privatize DOL under a threshold of 150 DOL to retain a public balance so they can pay future fees

---

## If PR to `staging`


tailwind.config.js
```diff
@@ -52,7 +52,8 @@ module.exports = {
         'button-fourth': 'var(--color-bg-button-fourth)',
         'connect-signer-button':'#2B49EA',
         'connect-wallet-button':'#00AFA5',
-        'gradient-button':'var(--gradient-button-secondary)'
+        'gradient-button':'var(--gradient-button-secondary)',
+        warning: '#FF8A0020'
```
zhangjing28, line 56: This color is not the same as the `warning` color in `textColor`. I think they should be the same

src/pages/SendPage/SendContext/index.tsx
```diff
@@ -314,16 +314,15 @@ export const SendContextProvider = (props) => {
   };
 
   // Returns true if the current tx would cause the user to go below a
-  // recommended min fee balance of 1. This helps prevent users from
+  // recommended min fee balance of 150. This helps prevent users from
   // accidentally becoming unable to transact because they cannot pay fees
   const txWouldDepleteSuggestedMinFeeBalance = () => {
```
zhangjing28, line 319: Nit: Since the result of `txWouldDepleteSuggestedMinFeeBalance` function has no argument that will influent the result of it, I think we could use a variable instead. The function will execute every time it calls, but a variable only executes once.

Boyuan-Deng (approved):
Agree with the above comments; otherwise, codes LGTM

tailwind.config.js
Kevingislason: These are the same shade with different opacity, I will call this `light-warning`

src/pages/SendPage/SendContext/index.tsx
Kevingislason: This code path runs exactly as often whether or not `txWouldDepleteSuggestedMinFeeBalance` is a function or a variable, since `txWouldDepleteSuggestedMinFeeBalance` is only called in one place. In similar cases where there is a notable performance difference, `useCallback` is best.

zhangjing28 (approved):

src/pages/SendPage/SendContext/index.tsx
zhangjing28: OK, get your point👍

PR #437, opened 2023-02-07
Feat/sbt generating

zhangjing28:
## Description

* generating page
* generated page

---

## If PR to `staging`

## If PR to `main`
- [ ] Update `CHANGELOG.md`


src/pages/SBTPage/SBTContext/sbtThemeContext.tsx
```diff
@@ -64,15 +64,22 @@ export const SBTThemeContextProvider = ({
...
     const ret = await axios.post<{
       model_id: string;
     }>(url, data);
     if (ret.status === 200 || ret.status === 201) {
       setModelId(ret?.data?.model_id);
     }
-  }, [checkedThemeItems, config, externalAccount, imgList]);
+  }, [
+    checkedThemeItems,
+    config.SBT_NODE_SERVICE,
```
zhangjing28, line 78: I tried to use `config` here, but I found it would be call this function again and again

src/pages/SBTPage/components/Generating/index.tsx
```diff
@@ -0,0 +1,75 @@
...
+              style={{
+                transform: transformStyle,
+                left,
+                zIndex: imgList.length - index,
+                opacity: 0.9
+              }}
+              key={index}
+            />
+          );
+        })}
```
Kevingislason, line 48: Whenever you have a `map` inside jsx, you can usually improve readability by factoring out the contents of the map into a subcomponent

src/pages/SBTPage/components/Generating/index.tsx
```diff
@@ -0,0 +1,75 @@
...
+  const { imgList } = useSBT();
+  const { queryGenerateResult } = useGenerating();
+
+  useEffect(() => {
+    const timer = setInterval(() => {
+      queryGenerateResult();
+    }, INTERVAL_TIME);
+
+    return () => clearInterval(timer);
+  }, [queryGenerateResult]);
```
Kevingislason, line 22: I think this counts as business logic, so it belongs upstream in `generatingContext`

src/pages/SBTPage/SBTContext/generatingContext.tsx
```diff
@@ -0,0 +1,80 @@
...
+  useContext,
+  useMemo,
+  useState
+} from 'react';
+
+import { useConfig } from 'contexts/configContext';
+import { useExternalAccount } from 'contexts/externalAccountContext';
+import { useSBTTheme } from './sbtThemeContext';
+
+type GenerateStatus = 'finish' | 'doing';
```
Kevingislason, line 15: This should probably just be an enum

src/pages/SBTPage/components/GeneratedImgs/index.tsx
```diff
@@ -0,0 +1,129 @@
...
+import { MAX_MINT_SIZE } from '../Generated';
+
+const PRE_SCALE = 0.07;
+const MAX_Z_INDEX = 30;
+
+type ItemType = {
+  imgUrl: string;
+  toggleMint: (imgUrl: string) => void;
+};
+const Item = ({ imgUrl, toggleMint }: ItemType) => {
```
Kevingislason, line 17: `Item` is a bit vague, it would be good to name this something more specific / descriptive

Kevingislason (commented):
One general comment about file structure: it seems like you have a lot of directories containing only one file. It will probably be easier to develop / navigate the SBT page code if you group files according to what step the files are used in.

src/pages/SBTPage/components/SignerButton/index.tsx
```diff
@@ -0,0 +1,44 @@
+import Icon from 'components/Icon';
```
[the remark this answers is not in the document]
zhangjing28: A similar reason to the `WalletButton`. But for this component, maybe we can extract a new Component like below, allowing the user to customize their display component in the future when we see another feature that will use this. For now, I think this duplication is just okay.
```javascript
const ZkAccountButton = ({customZkAccountDisplay}) => {
    const FinalZkAccountDisplay = customZkAccountDisplay || <ZkAccountDisplay />
    if (privateAddress) {
    return FinalZkAccountDisplay;
  } else if (signerIsOutOfDate(config, signerVersion)) {
    return (
      <ZkAccountWarning
        title={'Manta Signer Out of Date'}
        text={
          'Your Manta Signer install is out of date. Please install the latest version to view your zkAddress and zkAssets.'
        }
        showWarningIcon={true}
        showInstallButton={true}
      />
    );
  } else {
    return (
      <ZkAccountConnect
        className={
          'bg-connect-signer-button text-white h-12 w-36 font-medium cursor-pointer rounded-lg'
        }
      />
    );
  }
}
```

src/pages/SBTPage/components/WalletButton/index.tsx
```diff
@@ -0,0 +1,46 @@
+import ConnectWalletModal from 'components/Modal/connectWalletModal';
```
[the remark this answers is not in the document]
zhangjing28: I prefer to keep a little duplication. This `walletButton` is fit this special SBT feature, it's not used in another place. Modify the `components/Accounts/ConnectWallet` will plus its complexity and mix it with a special feature. If we have more features later, the `components/Accounts/ConnectWallet` might be full with `if(props.a)...if(props.b)...` and grow bigger and hard to maintain.

src/pages/SBTPage/SBTContext/generatingContext.tsx
zhangjing28: I think this is a little more convenient than an enum. For example, I need to import this enum anywhere I want to use it. But this `type` definition means it is a string first and it must be either `'finish'` or `'doing'`, any other string will not be allowed. I don't have to import any enum. Also, the IDE will tell me which value I can use like the image below. That's super cool!
image

src/pages/SBTPage/components/GeneratedImgs/index.tsx
zhangjing28: Done

src/pages/SBTPage/components/Generating/index.tsx
zhangjing28: Done

src/pages/SBTPage/components/Progress/index.tsx
```diff
@@ -0,0 +1,108 @@
...
+
+  useEffect(() => {
+    if (generateStatus === 'finish') {
+      setProgress(PROGRESS_MAX);
+      setRemainTime(0);
+      setTimeout(() => {
+        setCurrentStep(Step.Generated);
+      }, 100);
+    }
+  }, [generateStatus, setCurrentStep]);
```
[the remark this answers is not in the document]
zhangjing28: Done

src/pages/SBTPage/components/Progress/index.tsx
```diff
@@ -0,0 +1,108 @@
...
+    const timer = setInterval(() => {
+      if (remainTime <= REMAIN_TIME_BEFORE_FINISH) {
+        clearInterval(timer);
+      }
+      setRemainTime(remainTime - 1);
+    }, 60000);
+    return () => {
+      clearInterval(timer);
+    };
+  }, []);
```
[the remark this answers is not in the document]
zhangjing28: Done

src/pages/SBTPage/components/Progress/index.tsx
```diff
@@ -0,0 +1,108 @@
...
+
+const PROGRESS_MAX = 251.327;
+const PROGRESS_SPEED = 1;
+const PROGRESS_BEFORE_FINISH = 240;
+
+const REMAIN_TIME_BEFORE_FINISH = 5;
+
+const Progress = () => {
+  const [progress, setProgress] = useState(0);
+  const [remainTime, setRemainTime] = useState(20);
```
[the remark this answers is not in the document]
zhangjing28: Good idea!

src/pages/SBTPage/components/Progress/index.tsx
```diff
@@ -0,0 +1,108 @@
...
+            strokeDashoffset: `${progress}px`,
+            strokeLinecap: 'round'
+          }}></path>
+        <defs>
+          <linearGradient id="sbt-progress" gradientTransform="rotate(90)">
+            <stop offset="16.29%" stopColor="#2b49ea"></stop>
+            <stop offset="85.56%" stopColor="#00afa5"></stop>
+          </linearGradient>
+        </defs>
+      </svg>
```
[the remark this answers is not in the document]
zhangjing28: Done

src/pages/SBTPage/components/Generating/index.tsx
zhangjing28: `queryGenerateResult` is the business logic while the `setInterval` here is because I want to call the `queryGenerateResult` only in this `Generating` step, not the other steps. So I think it's okay to be here.

Kevingislason: Isn't there a variable for what step it is? You could just check the step variable in `generatingContext`

zhangjing28: Aha, that's a good idea, I didn't realize I can do that before.😂

src/pages/SBTPage/SBTContext/generatingContext.tsx
Kevingislason: I definitely think it's better to use an enum here. Enums are the preferred data type for a limited list of statuses, and importing them is a very small cost. Using strings for statuses like this can be confusing, because it is not clear from looking at strings that they represent one possible status, and it's not easy to find what other possible values are.

see: https://softwareengineering.stackexchange.com/questions/365339/what-is-wrong-with-magic-strings

zhangjing28: Okay, it definitely makes sense to me, I will update it.

Kevingislason (approved):

PR #439, opened 2023-02-09
fix conflicting tx result messages

Kevingislason:
## Description

closes: #387 

---

## If PR to `staging`

### DIFF
--- a/src/pages/SendPage/SendContext/index.tsx
+++ b/src/pages/SendPage/SendContext/index.tsx
@@ -401,6 +401,11 @@ export const SendContextProvider = (props) => {
         }
       }
     } else if (status.isFinalized) {
+      for (const event of events) {
+        if (api.events.utility.BatchInterrupted.is(event.event)) {
+          return;
+        }
+      }
       try {
         const signedBlock = await api.rpc.chain.getBlock(status.asFinalized);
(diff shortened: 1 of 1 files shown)


jincan39:
LGTM

src/pages/SendPage/SendContext/index.tsx
```diff
@@ -401,6 +401,11 @@ export const SendContextProvider = (props) => {
         }
       }
     } else if (status.isFinalized) {
+      for (const event of events) {
+        if (api.events.utility.BatchInterrupted.is(event.event)) {
+          return;
```
zhangjing28, line 406: Why not set `txStatus` to fail in this condition?

Kevingislason: The body of `handleTxRes` runs the first time when the transaction is put into a block, and a second time when the transaction's block is finalized. If a transaction fails, we will know that the transaction failed once it is in a block. So by the time a failed transaction is finalized, we have already handled the failure and surfaced it to the UI. So this code addition prevents the transaction from being handled twice.

zhangjing28: Got it

zhangjing28 (approved):

Issue #388, opened 2023-01-31
The balance of zkDol & zkKar are zero, but only zkDol showed on “zkAddress” 


bhgomes:

Kevingislason:
When we have a balance that doesn't fit, we want to use an elipsis like this:
![elipsis.png](https://images.zenhubusercontent.com/63d990be6ef44a1c5af59742/d430e414-4d89-4726-a63a-252629afc395)

Kevingislason:
Similarly, we want to show a tooltip on hover, design is at the bottom of this figma:
https://www.figma.com/file/9nAYHsYlfHmyT8RL5P8ISZ/Manta-Pay?node-id=0%3A1&t=Y2ZKz3WPXJwzepOC-0

PR #440, opened 2023-02-09
fix(style): privateBalance display 

jincan39:
## Description

fix privateBalance issue related to #388

---

## If PR to `staging`


jincan39:
image
image
If the balance not overflow it works as before and no tip show when mouse hover over text.
If the balance is text-overflow, the balance will be ellipsis and tip will be showed when mouse hover over text.

src/hooks/useEllipsis.tsx
```diff
@@ -0,0 +1,29 @@
+// @ts-nocheck
```
Kevingislason, line 1: Would you mind removing this and adding types where relevant? We are trying to slowly remove @ts-nocheck from the codebase, and to avoid it in new files.

src/hooks/useEllipsis.tsx
```diff
@@ -0,0 +1,29 @@
...
+import { useLayoutEffect, useState } from 'react';
+
+export const useEllipsis = (ref, callback) => {
+  const [isEllipsis, setIsEllipsis] = useState(undefined);
+
+  useLayoutEffect(() => {
```
Kevingislason, line 7: Is `useLayoutEffect` necessary? I actually haven't used this before, but the React docs say that plain `useEffect` is preferable for performance reasons. No worries if `useLayoutEffect` is necessary.

Kevingislason (commented):
Nice work on this, the useEllipsis hook is quite elegant

src/hooks/useEllipsis.tsx
jincan39: resolved

src/hooks/useEllipsis.tsx
jincan39: Good question. [useLayoutEffect](https://daveceddia.com/useeffect-vs-uselayouteffect/) will batch a couple updates together before redrawing the screen but it runs synchronously, in this scenario i think useEffect is enough to use. Btw, we should always thinking about batching and before React18 sometimes we consider [batching state updates](https://github.com/reactwg/react-18/discussions/21).

Kevingislason (approved):

src/hooks/useEllipsis.tsx
```diff
@@ -0,0 +1,30 @@
...
+      const isEllipsis =
+        offsetWidth && scrollWidth ? offsetWidth < scrollWidth : false;
+
+      current && setIsEllipsis(isEllipsis);
+      if (callback) callback(isEllipsis);
+    };
+
+    if (current) {
+      if ('ResizeObserver' in window) {
+        new ResizeObserver(trigger).observe(current);
```
zhangjing28, line 22: It's better add `unobserve` and `disconnect` logic as the returned result of `useEffect` here, check [this link](https://stackoverflow.com/questions/67581868/should-i-call-resizeobserver-unobserve-for-removed-elements) for further information

jincan39: Good notice, should avoid observer's leak.

src/hooks/useEllipsis.tsx
```diff
@@ -19,11 +22,15 @@ export const useEllipsis = (
...
-        new ResizeObserver(trigger).observe(current);
+        const observer: ResizeObserver = new ResizeObserver(trigger);
+        observer.observe(current);
+        observerRef.current = observer;
       }
 
       trigger();
     }
+
+    return observerRef.current.disconnect();
```
zhangjing28, line 33: The return of a `useEffect` function should be a callback function, which means it should be like this: 
```
    return () => {
        observerRef.current.disconnect()
    }

```

src/components/ZkAccount/PrivateAssetItem.tsx
```diff
@@ -0,0 +1,50 @@
+// @ts-nocheck
```
zhangjing28, line 1: Try to remove this.

src/styles/components.scss
```diff
@@ -122,6 +122,8 @@
 }
 
 
+// PrivateAssetItem
```
zhangjing28, line 125: Pls remove this

zhangjing28 (approved):

zhangjing28 (approved):

Kevingislason (approved):

Issue #401, opened 2023-01-31
After disconnect the account that is connecting, the address list not refreshed automatically


bhgomes:
2023.1.13: It will be better to give me a specific release version & deadline (Joly)
2023.1.12: Will fix, not priority

PR #442, opened 2023-02-10
detect address disconnection

Kevingislason:
## Description

closes: #401 

---

## If PR to `staging`


jincan39 (approved):

src/contexts/keyringContext.tsx
```diff
@@ -155,7 +150,8 @@ export const KeyringContextProvider = (props) => {
     if (!selectedWallet?.extension) {
       try {
         await selectedWallet.enable(APP_NAME);
-        subscribeWalletAccounts(selectedWallet, saveToStorage);
+        await refreshWalletAccounts(selectedWallet);
+        saveToStorage && setLastAccessedWallet(wallet);
```
zhangjing28, line 154: I think you mean `saveToStorage && setLastAccessedWallet(selectedWallet);`. The argument here should be `selectedWallet` but not `wallet`.

src/contexts/keyringContext.tsx
```diff
@@ -52,46 +52,41 @@ export const KeyringContextProvider = (props) => {
...
+
+      setSelectedWallet(wallet);
+      setKeyringAddresses(updatedAddresses);
     }
-    return unsubscribe;
+    keyringIsBusy.current = false;
   };
 
+  useEffect(() => {
+    const interval = setInterval(async () => {
```
zhangjing28, line 84: Just curious: Is there any way we can subscribe the change instead of using `setInterval`?

Kevingislason: Previously instead of this interval we used `wallet.subscribeAccounts`, but that missed the case I am attempted to solve in PR--the subscription does not detect it when the user revokes permission to connect to some address

zhangjing28: Got it, thanks for your explanation.

src/contexts/keyringContext.tsx
Kevingislason: Nice catch, thank you

PR #476, opened 2023-02-16
fix: upgrade the logic of handle transaction

zhangjing28:
## Description

* update the logic of handling transaction
* fix the bug of model_id is empty

---

## If PR to `staging`

## If PR to `main`
- [ ] Update `CHANGELOG.md`


src/pages/SBTPage/SBTContext/sbtPrivateWalletContext.tsx
```diff
@@ -136,11 +136,15 @@ export const SBTPrivateContextProvider = ({
             } else {
               console.error(error.toString());
             }
-            setTxStatus(TxStatus.failed(''));
```
Kevingislason, line 139: I think you may want to set `txStatus` to null here instead. Otherwise, `txStatus` will remain as 'Processing' even after the tx has finished.

src/pages/SBTPage/SBTContext/sbtPrivateWalletContext.tsx
```diff
@@ -136,11 +136,15 @@ export const SBTPrivateContextProvider = ({
...
             try {
-              const signedBlock: SignedBlock =
-                await sbtPrivateWallet.api.rpc.chain.getBlock(status.asInBlock);
+              const signedBlock: SignedBlock = status.isInBlock
+                ? await sbtPrivateWallet.api.rpc.chain.getBlock(
+                  status.asInBlock
+                )
+                : await sbtPrivateWallet.api.rpc.chain.getBlock(
+                  status.asFinalized
+                );
```
Kevingislason, line 147: What is the point of adding this case tbh?

Kevingislason (commented):
Many thanks for adding types `txStatusContext`

src/pages/SBTPage/SBTContext/sbtPrivateWalletContext.tsx
Kevingislason: Okay, sounds good. I thought you were trying to avoid showing a failed transaction notification.

src/pages/SBTPage/SBTContext/sbtPrivateWalletContext.tsx
```diff
@@ -118,7 +118,7 @@ export const SBTPrivateContextProvider = ({
       status: ExtrinsicStatus;
       events: EventRecord[];
     }) => {
-      if (status.isInBlock) {
+      if (status.isInBlock || status.isFinalized) {
```
Kevingislason, line 121: Why do you need to run this handler on `isFinalized`? The transaction is effectively complete as far as the user is concerned when it is `InBlock`. 

Also, this change may cause txStatus to be set more than once, which can cause strange behavior. For example, let's say you publish transaction A:  

1. transaction A is included in a block, so txStatus is set to Finalized. 
2. Next, you publish transaction B, and txStatus is set to Processing. 
3. You wait a while longer and transaction A is Finalized, so txStatus is set to Finalized. The user gets a second transaction success notification, even thought transaction B is still processing.

zhangjing28: I thought some tx might skip the `block` step😂。 Thanks for the explanation, will update it.

Kevingislason (approved):

Issue #397, opened 2023-01-31
show no message after install Talisman success but with no account


bhgomes:
2023.1.13: It will be better to give me a specific release version & deadline (Joly)
2023.1.12: Will fix

May require input from product team.

chloewang-p0x:
You have no account in Talisam. Please create one first.
https://www.figma.com/file/9nAYHsYlfHmyT8RL5P8ISZ/Manta-Pay?node-id=3365%3A658&t=CKZCDqTd0d8qIV8w-1

PR #483, opened 2023-02-16
feat: add talisman configure warning

jasonintju:
## Description

closes: #397 

---

## If PR to `staging`

## If PR to `main`
- [ ] Update `CHANGELOG.md`


jasonintju (commented):
Change the attribute of fill of every path to fill="currentColor", then svg will take the color of the font color, so we can change the svg color as we want. It's pretty cool.
I have checked all the components which using this svg, they all worked well.

Kevingislason (commented):
How does the app behave when we don't have a polkadot.js account of a SubWallet account? If possible, I think we should generalize this solution so it also handle cases where the user doesn't have a polkadot.js account or a SubWallet account.

jasonintju:
> How does the app behave when we don't have a polkadot.js account of a SubWallet account? If possible, I think we should generalize this solution so it also handle cases where the user doesn't have a polkadot.js account or a SubWallet account.

The other 3 wallets extension(Subwallet, polkadot.js, Metamask) behaved really well. When users click the 'Connect' button, if they don't have a account, the extensions will lead users to create a account first. Really good UX.

Only Talisman wallet needs to be configured first if users have no accounts. Because we didn't handle the Talisman connection error before, when users click the Talisman 'Connect' button, there is no feedback to the users, which is a bad UX. That's why only the Talisman was mentioned in this issue.

jincan39:
image
I experienced the keyring error above when I click the connect Tailsman wallet button.

jincan39:
lgtm

jincan39 (approved):

Issue #402, opened 2023-01-31
Don’t remember the account previous wallet selected


bhgomes:
2023.1.13: It will be better to give me a specific release version & deadline (Joly)
2023.1.12: Will fix, not priority

PR #493, opened 2023-02-17
fix: account previous wallet selected

jincan39:
## Description

closes: #402 

---

## If PR to `staging`


jincan39:
**First step**:  clear the chromium's localstorage (we use key "lastAccessedWallet" & "lastAccessedExternalAccountAddress")
**Second step**: connect polkadot wallet and select the other account , then refresh the browser to see if it can remember last address.
**Third step**: connect subwallet wallet and switch between polkadot wallet to see if the browser can remember last wallet with correct address previously selected.

src/contexts/keyringContext.tsx
```diff
@@ -52,46 +48,52 @@ export const KeyringContextProvider = (props) => {
...
+    keyringIsBusy.current = false;
+
+    if (reselected) {
+      const pairs = keyring.getPairs();
+      const {
+        meta: { source }
+      } = pairs[0] || { meta: {} };
+      const account =
+        getLastAccessedExternalAccount(keyring, source) || pairs[0];
+      return { account, pairs };
```
Kevingislason, line 86: This function was already very large / complex, so it's probably best to avoid giving it any more responsibility. So I think it might make sense to handle `reselected` logic, and returning `{ account, pairs }` in a different function. In places where currently you are calling `refreshWalletAccounts(wallet, true)`, you would do something like this: 

```js
await refreshWalletAccounts(wallet);
handleReselected(keyring) // or whatever function name you think is good
```

and places where you call `refreshWalletAccounts(wallet, true)` you can just call refreshWalletAccounts(wallet).

src/contexts/externalAccountContext.tsx
```diff
@@ -113,14 +113,19 @@ export const ExternalAccountContextProvider = (props) => {
         return;
       }
       // ensure newly added account after removing all accounts can be updated
```
Kevingislason, line 115: I don't think this comment still applies

Kevingislason (commented):
Nice work

src/utils/persistence/externalAccountStorage.ts
```diff
@@ -18,7 +18,12 @@ export const getLastAccessedExternalAccount = (config, keyring) => {
   }
 };
 
-export const setLastAccessedExternalAccountAddress = (config, lastAccount) => {
-  store.set(
-    `${config.BASE_STORAGE_KEY}${LAST_ACCOUNT_STORAGE_KEY}`, lastAccount);
+export const setLastAccessedExternalAccountAddress = (lastAccount) => {
+  const STORAGE_KEY = `${LAST_ACCOUNT_STORAGE_KEY}`;
```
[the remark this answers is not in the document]
jincan39: removed this

src/contexts/keyringContext.tsx
jincan39: nice catch, it's much more clear ^^

src/utils/persistence/externalAccountStorage.ts
```diff
@@ -3,10 +3,10 @@ import store from 'store';
 
 const LAST_ACCOUNT_STORAGE_KEY = 'lastAccessedExternalAccountAddress';
 
-export const getLastAccessedExternalAccount = (config, keyring) => {
-  const lastAccountAddress = store.get(
-    `${config.BASE_STORAGE_KEY}${LAST_ACCOUNT_STORAGE_KEY}`);
-
+export const getLastAccessedExternalAccount = ( keyring, walletType) => {
+  const STORAGE_KEY = `${LAST_ACCOUNT_STORAGE_KEY}`;
```
zhangjing28, line 7: I noticed we have a `src/constants/LocalStorageConstants.ts` file which includes some keys used for `localStorage`. So here are two questions: 

1. Should we define a key in that constants file or just define it in the storage file itself? 
2. Should we use the store package anymore since localStorage is supported by almost popular browsers? @Kevingislason

jincan39: Nice notice. It's a quite [outdated package](https://github.com/marcuswestin/store.js) since most people use [evergreen browsers](https://caniuse.com/?search=localstorage). I would suggest to gradually remove out the old package and introduce [zustand](https://github.com/pmndrs/zustand/discussions/524)(or some other elegant pkg) to manage the global store.

jasonintju: > I noticed we have a `src/constants/LocalStorageConstants.ts` file which includes some keys used for `localStorage`. So here are two questions:
> 
> 1. Should we define a key in that constants file or just define it in the storage file itself?
> 2. Should we use the store package anymore since localStorage is supported by almost popular browsers? @Kevingislason

I prefer defining the localStorage key in the constants file for an overall understanding of the keys we used in this project.

jasonintju (approved):

src/utils/persistence/externalAccountStorage.ts
jincan39: For the localstorage usage:
- [x] useLocalStorage()
- [ ] middleware [zustand-persist](https://www.npmjs.com/package/zustand-persist)
- [ ] middleware [rematch-persist](https://www.npmjs.com/package/@rematch/persist)
- [ ] etc
Firstly, i think to bring the useLocalStorage hook to replace the 'store' is ok which is not so heavy? and I'll try on another PR to figure out whether it works perfectlly @Kevingislason @zhangjing28

zhangjing28: I agree with using `useLocalStorage`

jincan39: Update notes: https://github.com/Manta-Network/manta-front-end/pull/565 (wip) useLocalStorage and persistence/storage.ts added which makes code more robustly. 
However, developers should still care about the storage's usage. To the state architecture and from personal perspective, something like [@rematch/persist](https://rematchjs.org/docs/plugins/persist/) which keep the same developer experience would be more charming and smell like perfume.
```    
const persistConfigs = {
      key: 'manta-web',
      storage: AsyncStorage,
      whitelist: ['localStorage','sender'],
    }
```

Kevingislason: Feel free to try using @rematch/persist if you think that approach is better

zhangjing28 (approved):

Issue #497, opened 2023-02-17
Border style in Analyzing page does not meet the design requirement


yingjingyang:
design draft
![截屏2023-02-17 下午10.28.20.png](https://images.zenhubusercontent.com/63e10f6258011a5593f8116e/884d3931-7c38-4ea1-9af2-c69bd622d443)

current website
![截屏2023-02-17 下午10.25.44.png](https://images.zenhubusercontent.com/63e10f6258011a5593f8116e/58416f5d-6178-4f97-b00f-cd6507ebbc6f)

Issue #538, opened 2023-02-20
After have selected 8 AI pictures, the other pictures should be gray


yingjingyang:
design draft
![截屏2023-02-20 下午7.44.21.png](https://images.zenhubusercontent.com/63e10f6258011a5593f8116e/e04fb44b-de01-4054-a0e4-67de345fa7d5)

current website
![截屏2023-02-20 下午7.44.13.png](https://images.zenhubusercontent.com/63e10f6258011a5593f8116e/bb4c1ce7-6aa0-4aad-8850-cc013f9710ab)

Issue #537, opened 2023-02-20
No need to show selected pictures in Button


yingjingyang:
design draft
![截屏2023-02-20 下午7.36.43.png](https://images.zenhubusercontent.com/63e10f6258011a5593f8116e/a296bb53-7d2e-49c4-b476-48d00145123f)

current website:
![截屏2023-02-20 下午7.36.05.png](https://images.zenhubusercontent.com/63e10f6258011a5593f8116e/80c74d93-f6e6-4595-9f72-05c4beda5eab)

Issue #535, opened 2023-02-20
In the second checkout window, when switch to another account, the error message is incorrect


yingjingyang:
design draft
![截屏2023-02-20 下午4.51.59.png](https://images.zenhubusercontent.com/63e10f6258011a5593f8116e/1e296d53-d9f3-4268-9f52-127c8977da49)

current website
![截屏2023-02-20 下午4.49.22.png](https://images.zenhubusercontent.com/63e10f6258011a5593f8116e/6f2daf4d-ff04-4262-9d38-4aa5fbd38c64)

PR #547, opened 2023-02-21
fix: some bugs related to theme, generating and generated page

zhangjing28:
## Description

closes: #497 #498 #502 #505 #503 #504 #506 #507 #508 #509 #512 #518 #519 #535 #537 #538 #549 #551 #552 #555 #556

---

- [X] Linked to Github issue with discussion and accepted design OR have an explanation in the PR that describes this work
- [X] Re-reviewed files changed in the Github PR explorer

## If PR to `staging`

## If PR to `main`
- [ ] Update `CHANGELOG.md`


src/hooks/useModal.tsx
```diff
@@ -79,13 +84,13 @@ export const useModal: (options?: IUseModal) => any = (
   }, [open]);
 
   const ModalWrapper = useMemo(
-    () =>
-      ({ children }: { children: React.ReactNode }) => {
+    function ModalWrapper() {
```
zhangjing28, line 87: Just add a function name to avoid the eslint error

src/pages/SBTPage/components/Generating/index.tsx
```diff
@@ -51,24 +75,35 @@ const Generating = () => {
...
+            </p>
+          ) : (
+            <>
+              <p className="text-xl">While you are waiting...</p>
+              <div className="text-warning rounded-xl py-1 my-4 flex">
+                <Icon name="information" className="mr-2" /> You will need Manta
+                Signer to mint your SBTs
+              </div>
+              <p className="text-white text-opacity-60">
+                Have not signed in your Manta Wallet/Signer yet? <br />
```
Kevingislason, line 93: I think this should be "Have you signed in your Manta Wallet/Signer yet?"

src/pages/SBTPage/components/Progress/index.tsx
```diff
@@ -8,14 +9,49 @@ import {
...
+const REMAIN_TIME_BEFORE_FINISH = 2;
+const REMAIN_TIME_MAX = 20;
+
+enum initialType {
+  time,
+  progress
+}
+
+// count the time and progress even the user left the page
+const initialProgressAndRemainTime = (type: initialType) => {
```
Kevingislason: I think this parameter is a little confusing and this function is a bit complex. It would probably be best to split this function into two functions: one which returns time, one which returns progress. If you don't think there is a good way to do that, you could also return `{progressResult, timeResult}`.

src/types/Balance.ts
```diff
@@ -61,7 +65,7 @@ export default class Balance {
       .toNumber()
       .toLocaleString(undefined, {
         maximumFractionDigits: decimals,
-        minimumFractionDigits: 0,
+        minimumFractionDigits: decimals
```
Kevingislason, line 68: Not sure we want to do this, as this will change balance display on all other front end pages.

src/pages/SBTPage/components/MintCheckModal/index.tsx
```diff
@@ -124,7 +124,7 @@ const MintCheckModal = ({
 
   const errorMsg = useMemo(() => {
     if (generateAccount?.address !== externalAccount?.address) {
-      return 'The address to mint must be the same as the address to genereate.';
+      return 'Please switch to the same wallet/network that you made the purchase for the generation';
```
Kevingislason, line 130: I think this should be something 'Please switch to the same wallet/network that you used to make the purchase for the generation'

src/pages/SBTPage/components/MintCheckModal/index.tsx
```diff
@@ -172,7 +179,7 @@ const MintCheckModal = ({
         Balance: {nativeTokenBalance?.toDisplayString() ?? '-'}
       </p>
       {errorMsg && (
-        <p className="mt-2 text-error text-left">
+        <p className={`${errorMsgStyle} mt-2 text-left text-tip`}>
```
Kevingislason: I think you will have `text-tip` in your `className` twice if `errorMsgStyle` is also `text-tip`

src/pages/SBTPage/SBTContext/sbtPrivateWalletContext.tsx
```diff
@@ -216,25 +214,40 @@ export const SBTPrivateContextProvider = ({
...
+        const proofId = toHexString(
+          tx[0].ToPrivate[0]['utxo_commitment_randomness']
+        );
+        const assetId = new BN(tx[0].ToPrivate[1]['id'], 'le').toString();
+        return {
+          assetId,
+          proofId
+        };
+      });
+      return ret;
```
Kevingislason, line 236: Instead of assigning `transactionDatas.map(...)` to `res`, I think you can just directly `return transactionDatas.map(...)`

src/utils/Unit8Array/toHexString.ts
```diff
@@ -0,0 +1,6 @@
+export function toHexString(byteArray: Uint8Array) {
+  return byteArray.reduce(
+    (output, elem) => output + ('0' + elem.toString(16)).slice(-2),
+    ''
+  );
+}
```
Kevingislason, line 6: I think you can just use this `import { u8aToHex } from '@polkadot/util`'

src/pages/SBTPage/components/Generating/index.tsx
zhangjing28: Thanks for your professional advice, will update that.

src/pages/SBTPage/components/Progress/index.tsx
zhangjing28: Okay, I will create two functions to reduce the complexity of the function.

src/types/Balance.ts
zhangjing28: I think we can do this, this just makes the number like `178.0` to `178.000` if the decimal is three. It's better we keep all the numbers that have the same decimal.

src/pages/SBTPage/components/MintCheckModal/index.tsx
zhangjing28: Thanks for that, will do

src/pages/SBTPage/components/MintCheckModal/index.tsx
zhangjing28: Good catch, that's a mistake, will fix it

src/pages/SBTPage/SBTContext/sbtPrivateWalletContext.tsx
zhangjing28: Good idea, will do

src/utils/Unit8Array/toHexString.ts
zhangjing28: Happy to know that, will try it

src/pages/SBTPage/SBTContext/sbtPrivateWalletContext.tsx
```diff
@@ -79,10 +82,8 @@ export const SBTPrivateContextProvider = ({
       const sbtPrivateWallet = await SbtMantaPrivateWallet.initSBT(
         privateWalletConfig
       );
-      console.log('init sbtPrivateWallet after', sbtPrivateWallet);
-
-      await sbtPrivateWallet.initalWalletSync();
       setSBTPrivateWallet(sbtPrivateWallet);
+      console.log('init sbtPrivateWallet after', sbtPrivateWallet);
```
Kevingislason, line 86: Might want to get rid of this, or add comment if you want to keep it

src/pages/SBTPage/components/MintedModal/index.tsx
```diff
@@ -1,14 +1,18 @@
+//@ts-nocheck
```
Kevingislason, line 1: I should have caught this earlier, but you will probably want to avoid this. If you are having type problems, you can disable type checking for a single line or use the `any` type

src/pages/SBTPage/components/Progress/index.tsx
```diff
@@ -8,14 +9,66 @@ import {
...
+  const pastMin = (now - generatingStart) / 1000 / 60;
+  if (!generatingStart) {
+    timeResult = REMAIN_TIME_MAX;
+  } else if (pastMin > REMAIN_TIME_MAX) {
+    timeResult = REMAIN_TIME_BEFORE_FINISH;
+  } else {
+    timeResult = REMAIN_TIME_MAX - Math.floor(pastMin);
+  }
+  return timeResult;
+};
```
Kevingislason, line 59: There's quite a bit of duplication between these two functions, maybe you could factor out the shared logic into a third function.

src/types/Balance.ts
Kevingislason: Other crypto apps don't usually show more decimals than necessary, it takes up space on the UI without providing additional information. The MantaPay design doesn't use additional decimals. I'm not sure why you would want to do this tbh. Is it part of the design for sbt? If it's truly necessary, I'd rather you write another toString function than modify the one used elsewhere in the app.

src/pages/SBTPage/SBTContext/sbtPrivateWalletContext.tsx
zhangjing28: Will get rid of it.

src/pages/SBTPage/components/Progress/index.tsx
zhangjing28: Yeah, I noticed that too. But to be honest, I have no idea how to do that. Do you have any advice about that?

src/pages/SBTPage/components/MintedModal/index.tsx
zhangjing28: This was added because of the `Popover` component in `element-react`, and I tried to get rid of it, but I failed. Let me give it another try

src/types/Balance.ts
zhangjing28: I want to do this because the function before makes the number  '178' to be '178.0', and '178.1234' to be '178.123'. It will keep a number have a different decimal. I just add this to make all the numbers have the same decimal whether it's '178' or '178.1234'

src/pages/SBTPage/components/MintedModal/index.tsx
zhangjing28: Removed

src/pages/SBTPage/components/Progress/index.tsx
Kevingislason: Something like this, although I don't think the function name is good
```js
  const getGeneratingStart = (now: Date) => {
       const finalKey = `${localStorageKeys.GeneratingStart}-${externalAddress}`;
       const generatingStart = Number(localStorage.getItem(finalKey));
      if (!generatingStart || now < generatingStart) {
         localStorage.setItem(finalKey, `${now}`);
      }
  }

```

Kevingislason (approved):

PR #565, opened 2023-02-21
refactor: useLocalStorage to replace store.js

jincan39:
## Description

closes: #XXXX

---

## If PR to `staging`


jasonintju (commented):
I think we'd better not replace all the codes in only one PR. It needs a lot of tests.
It will be better if we use the uselocalstorage for new codes and gradually update the old ones. 
Btw, this repository code recipes are pretty clear. https://github.com/uidotdev/usehooks

src/hooks/useIsomorphicLayoutEffect.ts
```diff
@@ -0,0 +1,6 @@
+import { useEffect, useLayoutEffect } from 'react';
+
+const useIsomorphicLayoutEffect =
```
zhangjing28, line 3: Why do we have to use this?

src/hooks/useLocalStorage/useLocalStorage.ts
```diff
@@ -0,0 +1,103 @@
...
+    (event: StorageEvent | CustomEvent) => {
+      if ((event as StorageEvent)?.key && (event as StorageEvent).key !== key) {
+        return;
+      }
+      setStoredValue(readValue());
+    },
+    [key, readValue]
+  );
+
+  // this only works for other documents, not the current one
```
zhangjing28, line 83: It's not a proper annotation

src/hooks/useLocalStorage/useLocalStorage.ts
```diff
@@ -0,0 +1,103 @@
...
+  useEventListener('storage', handleStorageChange);
+
+  // this is a custom event, triggered in writeValueToLocalStorage
```
zhangjing28, line 86: It's not a proper annotation, So as the next annotation

src/utils/persistence/storage.ts
```diff
@@ -0,0 +1,52 @@
...
+      entry = window.localStorage.getItem(key);
+    } catch (error) {
+      console.error('getFromLocalStorage', error);
+      return undefined;
+    }
+
+    try {
+      entry = JSON.parse(entry);
+    } catch (e) {
+      // Just use raw value
```
zhangjing28, line 39: It's better add `console.error(e)` and remove the annotation here.

src/utils/persistence/storage.ts
```diff
@@ -0,0 +1,52 @@
...
+  try {
+    window.localStorage.removeItem(key);
+  } catch (error) {
+    console.error('removeLocalStorage', error);
+  }
+}
+
+export function getFromLocalStorage(key: string) {
+  if (typeof window !== 'undefined' && window.localStorage) {
+    let entry: any;
```
zhangjing28, line 28: Try to use `string | null | undefined` instead of any

src/utils/persistence/storage.ts
```diff
@@ -0,0 +1,52 @@
...
+    try {
+      entry = window.localStorage.getItem(key);
+    } catch (error) {
+      console.error('getFromLocalStorage', error);
+      return undefined;
+    }
+
+    try {
+      entry = JSON.parse(entry);
```
zhangjing28, line 37: I think it's another Object after calling `JSON.parse(entry)`. You should define another variable for this.

zhangjing28 (commented):
Agree with Jason, we should test the functions effected by this refactor

src/hooks/useIsomorphicLayoutEffect.ts
jincan39: Good question. We can simply use useEffect currently since we don't need hydrate components. Just [delay show component](https://gist.github.com/gaearon/e7d97cdf38a2907924ea12e4ebdf3c85#option-2-lazily-show-component-with-uselayouteffect) to avoid some UI broken cases when the project use server rendering.

jincan39:
> I think we'd better not replace all the codes in only one PR. It needs a lot of tests. It will be better if we use the uselocalstorage for new codes and gradually update the old ones. Btw, this repository code recipes looks pretty clear. https://github.com/uidotdev/usehooks

For the consideration, I can split into different PRs and this recipe is quite a straight one. However, we'll have two different way to handle localstorage (useLocalStorage & store.js) which would confuse developers. After a closer glance at store.js, I'd prefer to turn this PR into draft at present and let's discuss later after all the bug issues finished.

jasonintju:
After using store.js today, I'm kinda loving it.

zhangjing28:
In that case, I think we can convert this PR into a draft.

Issue #598, opened 2023-02-22
The zkAddress style need to be consistent with the design draft


yingjingyang:
design draft 

![截屏2023-02-22 下午11.09.46.png](https://images.zenhubusercontent.com/63e10f6258011a5593f8116e/d9c360d8-0653-49ee-8d14-8b1f349258ed)

current website

![截屏2023-02-22 下午11.12.43.png](https://images.zenhubusercontent.com/63e10f6258011a5593f8116e/347308c0-253a-4087-929a-0ce25a9f27fa)

PR #613, opened 2023-02-23
style: CopyPasteIcon component

jasonintju:
## Description

- closes: #598 
- refactor the CopyPasteIcon component
- delete some unused files
---

## If PR to `staging`

## If PR to `main`
- [ ] Update `CHANGELOG.md`


jasonintju:
> Code looks good, thanks also for getting rid of these unused files. Just a few small style tweaks:
> 
> 1. The corners around the copy button in the design seem a bit more rounded
> 2. The copy button's background color should match the design, currently it's a bit darker.
> 3. The copy button and the zkAddress display have a bit more space between them in the design.

I think you will be a master of the UI design, with really goooood insight and carfulness. That's really impressive.
Sorry for the 1 & 2, it's my mistake, I'll fix it right now. But for the 3, I don't know what the final design draft looks like. What I got is looks like this. It is different with the UI we are working on, so I can not tell the exactly space between the copy button and the zkAdress display.

image

Kevingislason:
I'm looking at the same mockup you are tbh. I don't know what exact distance to use, but the borders around the two components are just a few pixels apart, seems like there is slightly more space in the mockup. Maybe try adding 0.5rem and seeing how it looks.

jasonintju:
> I'm looking at the same mockup you are tbh. I don't know what exact distance to use, but the borders around the two components are just a few pixels apart currently. It seems like there is slightly more space in the mockup. Maybe try adding 0.5rem and seeing how it looks.

Yeah, the space is about 8px, it's not OK?
The problem is that I can't add the space myself. The space is decided by the icon button's width and the zkAddress display's width. 'flex justify-between' classNames were used for their parent div. The button's width is fixed, and the zkAddress display's width is almost fixed too. So, if we want more space, we have to change the zkAddress display styles, such as the gap, the address's length and more. I think that's not what we want.

image

src/components/ZkAccount/ZkAccountModal.tsx
```diff
@@ -55,14 +55,14 @@ const ZkAddressDisplay = () => {
...
-      <div className="flex items-center gap-2">
+    <div className="flex justify-between">
+      <div className="border border-secondary bg-white bg-opacity-5 rounded-lg p-2 flex items-center gap-2">
         <Icon className="w-6 h-6" name="manta" />
         <span className="text-white font-light">
           {privateAddressDisplayString}
         </span>
       </div>
-      <CopyPasteIcon className="w-5 h-5" textToCopy={privateAddress} />
+      <CopyPasteIcon btnClassName="border border-secondary bg-white bg-opacity-5 rounded-lg relative p-2.5 w-10.5 text-0" iconClassName="w-5 h-5" textToCopy={privateAddress} />
```
jincan39, line 65: Generally it looks good to me. should we use something like w-[42px]? the different unit 'px' & 'rem''s usage and we extend tailwind's design-system casually kind of confusing to me.

tailwind.config.js
```diff
@@ -3,6 +3,7 @@ module.exports = {
   darkMode: 'class',
   theme: {
     fontSize: {
+      0: '0px',
```
jincan39, line 6: should we use '0rem' here as the design-system's basic unit is rem

jincan39 (approved):

tailwind.config.js
jasonintju: 0 rem = 0 * fontSize = 0 px
so, it does't make any sense. Just leave it.

src/components/ZkAccount/ZkAccountModal.tsx
jasonintju: > Generally it looks good to me. should we use something like w-[42px]? the different unit 'px' & 'rem''s usage and we extend tailwind's design-system casually kind of confusing to me.

v2 doesn't support w-[42px]

Kevingislason:
I still think it looks kind of weird tbh. You are always free to mess with the existing styles slightly as long as things basically look good. One thing I would try is slightly shrinking the font. I think removing one letter from the address display would be fine too.

jasonintju:
> You are always free to mess with the existing styles slightly as long as things basically look good

Wow, that's great. I will showcase my talent in UI design.

jasonintju:
Removing one letter seems good. @Kevingislason 
image

Kevingislason (approved):
Looks good, nice work

Issue #448, opened 2023-02-14
When the zkAsset is synchronization, the zkAddress sub-window should potentially show the sync is happening state


yingjingyang:

jasonintju:
The issue was talking about this?
![manta-issue-pic.png](https://images.zenhubusercontent.com/63eb210f44616d2183a83ce2/b7491201-41ff-4cac-a5f7-fb55f187bd25)

PR #610, opened 2023-02-23
fix: send button status

jincan39:
## Description

closes: #604

---

Before we can merge this PR, please make sure that all the following items have been checked off. If any of the checklist items are not applicable, please leave them but write a little note why.

- [x] Linked to Github issue with discussion and accepted design OR have an explanation in the PR that describes this work
- [x] Re-reviewed files changed in the Github PR explorer

## If PR to `staging`
- [x] Updated relevant documentation in the code
- [x] Test code changes manually, and describe your procedure in a comment on this PR
- [x] Mark this PR with the correct milestone

PR #625, opened 2023-02-24
fix: zkAsset sync tip

jincan39:
## Description

closes: #448

---

## If PR to `staging`

### DIFF
--- a/src/components/ZkAccount/PrivateAssetTableContent.tsx
+++ b/src/components/ZkAccount/PrivateAssetTableContent.tsx
@@ -3,17 +3,29 @@ import {
   useZkAccountBalances,
   ZkAccountBalance
 } from 'contexts/zkAccountBalancesContext';
+import { useSend } from 'pages/SendPage/SendContext';
 import PrivateAssetItem from './PrivateAssetItem';
 
 const PrivateAssetTableContent = () => {
   const { balances } = useZkAccountBalances();
+  const { senderIsPrivate, receiverIsPrivate } = useSend();
   const privateWallet = usePrivateWallet();
+  const isInitialSync = privateWallet?.isInitialSync?.current;
+  const shouldSyncZKAccount =
+    (isInitialSync && senderIsPrivate()) || (isInitialSync && receiverIsPrivate());
+
   if (balances?.length) {
     return (
       <div className="divide-y divide-dashed divide-manta-gray-secondary">
         {balances.map((balance: ZkAccountBalance) => (
           <PrivateAssetItem balance={balance} key={balance.assetType.assetId} />
         ))}
+      </div>
+    );
+  } else if (shouldSyncZKAccount) {
+    return (
+      <div className="whitespace-nowrap text-center mt-6">
+        Syncing zk account
       </div>
     );
   } else if (privateWallet?.balancesAreStaleRef.current) {
--- a/src/components/ZkAccount/PrivateAssetTableContent.tsx
+++ b/src/components/ZkAccount/PrivateAssetTableContent.tsx
@@ -3,29 +3,33 @@ import {
   useZkAccountBalances,
   ZkAccountBalance
 } from 'contexts/zkAccountBalancesContext';
-import { useSend } from 'pages/SendPage/SendContext';
+import useReceiverBalanceText from 'pages/SendPage/SendToForm/useReceiverBalanceText';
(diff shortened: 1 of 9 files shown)


src/components/ZkAccount/PrivateAssetTableContent.tsx
```diff
@@ -3,11 +3,17 @@ import {
...
+import { useSend } from 'pages/SendPage/SendContext';
 import PrivateAssetItem from './PrivateAssetItem';
 
 const PrivateAssetTableContent = () => {
   const { balances } = useZkAccountBalances();
+  const { senderIsPrivate, receiverIsPrivate } = useSend();
   const privateWallet = usePrivateWallet();
+  const isInitialSync = privateWallet?.isInitialSync?.current;
+  const shouldSyncZKAccount =
+    (isInitialSync && senderIsPrivate()) || (isInitialSync && receiverIsPrivate());
```
jasonintju, line 15: How about `isInitialSync && (senderIsPrivate() || receiverIsPrivate())`?

src/components/ZkAccount/PrivateAssetTableContent.tsx
```diff
@@ -3,11 +3,17 @@ import {
...
 } from 'contexts/zkAccountBalancesContext';
+import { useSend } from 'pages/SendPage/SendContext';
 import PrivateAssetItem from './PrivateAssetItem';
 
 const PrivateAssetTableContent = () => {
   const { balances } = useZkAccountBalances();
+  const { senderIsPrivate, receiverIsPrivate } = useSend();
   const privateWallet = usePrivateWallet();
+  const isInitialSync = privateWallet?.isInitialSync?.current;
+  const shouldSyncZKAccount =
```
Kevingislason, line 14: I don't think it makes sense to check `senderIsPrivate() || receiverIsPrivate()` here. During the initial sync, we don't have enough information to show zkAccount balances no matter what. It doesn't matter if the selected asset types in the main form are private or not.

jincan39: Make sense, think we should check like `(receiverIsPrivate && receiverLoading) ||
    (senderIsPrivate && senderLoading).` rebased on #610

src/components/ZkAccount/PrivateAssetTableContent.tsx
```diff
@@ -3,12 +3,28 @@ import {
...
 import PrivateAssetItem from './PrivateAssetItem';
 
 const PrivateAssetTableContent = () => {
   const { balances } = useZkAccountBalances();
   const privateWallet = usePrivateWallet();
-  if (balances?.length) {
+  const { shouldShowLoader: receiverLoading, receiverIsPrivate } =
+    useReceiverBalanceText();
+  const { shouldShowLoader: senderLoading, senderIsPrivate } =
+    useSenderBalanceText();
```
Kevingislason, line 16: Sorry, I should have given this suggestion in my previous review, but I actually don't think this component should depend on  `useSenderBalanceText` or `useReceiverBalanceText`. These two hooks are in charge of display logic for specific components in the `SendPage` specifically. That logic does not directly apply here. Also `PrivateAssetTableContent` does not contain any information about receiver balances.

src/components/ZkAccount/PrivateAssetTableContent.tsx
Kevingislason: I think the check you want is just `isInitialSync.current`. You can add this to the check on line 35 instead of adding a new `if` condition:
`} else if (privateWallet?.balancesAreStaleRef.current || isInitialSync.current) {`

src/components/ZkAccount/PrivateAssetTableContent.tsx
jincan39: Appreciated, it's not your fault. I rebased the code so it got refreshed 😂

src/components/ZkAccount/PrivateAssetTableContent.tsx
jincan39: Understand, yeah it's a much simpler approach. But it not works because `balances?.length` always return true, let me configure it out.

jincan39: I opened another pr to fix the initialSync logic issue #637

Issue #600, opened 2023-02-23
Don't need to show "ongoing task" when user have already minted


yingjingyang:
![截屏2023-02-23 上午10.02.48.png](https://images.zenhubusercontent.com/63e10f6258011a5593f8116e/41561428-6fd5-42ae-9fcc-5e8cb3916640)

PR #656, opened 2023-03-03
fix: ongoing notification when skip step

jincan39:
## Description

closes: #600

---

## If PR to `staging`


src/pages/SBTPage/components/OnGoingTaskModal/index.tsx
```diff
@@ -1,13 +1,17 @@
...
+import { useSBTTheme } from 'pages/SBTPage/SBTContext/sbtThemeContext';
 import { useEffect } from 'react';
 
 const OnGoingTaskNotification = () => {
   const { showOnGoingTask, setCurrentStep, onGoingTask, toggleSkippedStep } =
     useSBT();
+  const { modelId } = useSBTTheme();
 
   useEffect(() => {
+    if (modelId && onGoingTask?.model_id === modelId) return; // skippedStep
```
zhangjing28, line 13: To be honest, I don't understand why add this logic?

jincan39: The bug happened when user refresh the page, he do have onGoingTask and waiting for it to be finished. When the task is finished we should check whether user jump from the generating page. 
 - firstly the `mode_id` was set in two places from the code: 1. after `generateImgs` 2. skipped step
 - with a normal process the bug won't happen because when user navigate back, the `onGoingTask` always keeps empty.
 - if user refresh the page, `onGoingTask` will be set initially and  the `model_id ` won't set immediately but after user skip steps to the generating page. When user click `Go to home page`, we should keep safe check whether user match the case that he did refreshed page and skipped steps.

jincan39: fixed by normal approach

src/pages/SBTPage/SBTContext/mintContext.tsx
```diff
@@ -66,6 +66,7 @@ export const MintContextProvider = ({ children }: { children: ReactNode }) => {
 
   const resetContextData = useCallback(() => {
     setImgList([]);
+    setOnGoingTask({ status: true, model_id: '', urls: [] });
```
zhangjing28, line 69: I think just `setOnGoingTask(null)` is fine

zhangjing28 (approved):

PR #875, opened 2023-04-11
update manta wallet select dropdown

jasonintju:
## Description
image

---

## If PR to `staging`

## If PR to `main`
- [ ] Update `CHANGELOG.md`


jincan39 (approved):

src/components/Accounts/AccountSelectDropdown.tsx
```diff
@@ -39,6 +40,59 @@ const SingleAccountDisplay = ({
...
+                textToCopy={accountAddress}
+              />
+            </div>
+          </div>
+          <div className="relative right-2">
+            <Icon name="greenCheck" />
+          </div>
+        </div>
+        
+      </div>
```
Kevingislason, line 93: It might be good to factor out two subcomponents, one for the private address, one for the public address

Kevingislason (commented):
UI looks good. The one thing remaining here is to prevent the manta wallet private address from appearing when the user is using the old version of the front end that supports manta signer instead.

jasonintju:
> UI looks good. The one thing remaining here is to prevent the manta wallet private address from appearing when the user is using the old version of the front end that supports manta signer instead.

not only private address, but the whole manta wallet. Chloey said users won't see manta wallet in the old version. They're totally two different versions, no any relationship.

src/components/Accounts/AccountSelectDropdown.tsx
jasonintju: pls review again, thanks

Kevingislason (approved):

PR #881, opened 2023-04-12
refactor: keyring init logic

jincan39:
## Description

fix lastAccess wallet should be remembered after user refresh page.

---

## If PR to `staging`

## If PR to `main`
- [ ] Update `CHANGELOG.md`


src/contexts/keyringContext.tsx
```diff
@@ -85,6 +88,7 @@ export const KeyringContextProvider = ({
   );
 
   const refreshWalletAccounts = async (wallet: Wallet) => {
+    if (!wallet?.enable) return Promise.resolve('no enable function'); // safe check
```
Kevingislason, line 91: nit: one style convention we've maintained throughout the repo is to avoid inline conditionals, so something like this would be a little better:
```js
if (!wallet?.enable) {
  return Promise.resolve('no enable function')
}
```

src/contexts/keyringContext.tsx
```diff
@@ -118,6 +122,7 @@ export const KeyringContextProvider = ({
       setKeyringAddresses(substrateAddresses);
     }
 
+    if (wallet.extensionName === WALLET_NAME.MANTA) setSelectedWallet(wallet); // TODO manta zkAsset, move logic outside
```
Kevingislason, line 125: nit: one style convention we've maintained throughout the repo is to avoid inline conditionals, so something like this would be a little better:
```js
if (wallet.extensionName === WALLET_NAME.MANTA) {
  setSelectedWallet(wallet)
}
```

Kevingislason, line 125: Can you explain what you mean by the todo comment?

src/contexts/keyringContext.tsx
```diff
@@ -178,8 +183,8 @@ export const KeyringContextProvider = ({
 
   const connectWallet = useCallback(
     async (extensionName: string, saveToStorage = true) => {
-      if (!isKeyringInit || !extensionName) {
-        return;
+      if (!isKeyringInit) {
+        return Promise.resolve('keyring not init');
```
Kevingislason, line 187: There is code in `components/modal/connectWalletModal` that expects `connectWallet` to return a boolean. I think you may need to modify that code.

Kevingislason (commented):
Seems to fix the problem, nice work!

src/contexts/keyringContext.tsx
jincan39: Nice catch

src/contexts/keyringContext.tsx
jincan39: Think i met problem when switch signer to wallet, i will remove this unclear line of code first.

jasonintju (approved):

PR #944, opened 2023-04-20
fix: fetchZkBalances usage

jincan39:
## Description
Almost same story to #948, we should move out the getSpendableBalance method from interval.

1、When to refetch the zkBalance for Manta Wallet

    1、initialized, watch isWalletReady  to become true 
    2、switch token(ticker) 
    3、after transaction isFinalized and privateWallet walletSync is done.
    4、fallback, every 2 minutes for the active tab

---

## If PR to `staging`


src/pages/SendPage/SendContext/index.tsx
```diff
@@ -287,14 +294,43 @@ export const SendContextProvider = (props) => {
...
+    }, 120000); // refetch every two mins
+    return () => clearInterval(interval);
+  }, []);
+
+  useEffect(() => {
+    const interval = setInterval(() => {
+      if (txStatus?.isProcessing()) {
+        return;
+      }
+      fetchFeeBalance(); // TODO
```
Kevingislason, line 333: You may want to resolve this todo. It seems like you would want to fetch fee balances whenever you fetch the other balances, rather than every second.

jincan39: think fee is related to zkBalance, i just put this method into fetchZkBalances. don't whether it is the right way to do

Kevingislason (approved):

PR #981, opened 2023-04-26
Feat/add erc stablecoins

Kevingislason:
## Description

- Enables bridging of USDC, USDT, and DAI between Karura and Calamari
- Enables bridging of USDT between Statemine and Calamari
- Enables USDC, USDT, and DAI on MantaPay

---

## If PR to `main`
- [x] Update the version number in `package.json` and `src/config/common.json`
- [x] Update the minimum required signer version number in `src/config/common.json` if applicable
- [x] Update `CHANGELOG.md`
- [x] Make sure that all PR's to `manta-signer` and `sdk` on which this PR depends are merged


patches/@acala-network+sdk+4.1.8-14.patch
```diff
@@ -0,0 +1,32 @@
+diff --git a/node_modules/@acala-network/sdk/wallet/balance-adapter/erc20-adapter.js b/node_modules/@acala-network/sdk/wallet/balance-adapter/erc20-adapter.js
```
Kevingislason, line 1: This .patch file plus `patch-package` are the standard way of patching npm dependencies. This patch solves a problem causing balances of acala erc tokens not to update by getting rid of some caching.

src/types/Chain.ts
```diff
@@ -59,21 +60,6 @@ export default class Chain {
     this.api = null;
   }
 
-  static Dolphin(config) {
```
jasonintju, line 62: Why deleted Dolphin? It will caused error when changed to Dolphin network.

jasonintju:
I'm not sure why I can't see DAI and USDC in the selector.

jasonintju (dismissed):
Code looks good. I'll do more tests once I get the test tokens.

src/types/Chain.ts
Kevingislason: Dolphin runtime isn't used anymore. The testnet we call Dolphin is called "DolphinSkinnedCalamari" in this file because it's a copy of Calamari runtime behind the scenes, but we treat it as "Dolphin" in the UI. Does that make sense?

jincan39 (approved):
lgtm

src/types/Chain.ts
jasonintju: Yeah, it makes sense. 
It's just a little weird that there are errors throw out when I run the `feat/add_erc_stablecoins` branch locally. But in vercel, it seems ok.
image

Kevingislason:
I removed Karura from the Dolphin bridge page since none of these assets are registered, so the XCM adapter errors out (even for KAR)

jasonintju (dismissed):

jincan39 (approved):

PR #1007, opened 2023-05-06
fix: initial gas fee & insufficient balance

jincan39:
## Description

update the gas fee if user make transaction and mention users to reserve some KMA. 

---

## If PR to `main`
- [ ] Update `CHANGELOG.md`

### DIFF
--- a/src/contexts/globalContexts.tsx
+++ b/src/contexts/globalContexts.tsx
@@ -19,6 +19,7 @@ const GlobalContextProvider = ({ children }) => {
   const initValue = store.get(localStorageKeys.UsingMantaWallet, true);
   const [usingMantaWallet, _setUsingMantaWallet] = useState(initValue);
   const [mantaWalletInitialSync, _setMantaWalletInitialSync] = useState(true);
+  const [suggestedMinFeeBalance, setSuggestedMinFeeBalance] = useState();
   const { pathname } = useLocation();
   const isDolphinPage = pathname.includes('dolphin');
 
@@ -44,13 +45,17 @@ const GlobalContextProvider = ({ children }) => {
       usingMantaWallet,
       setUsingMantaWallet,
       mantaWalletInitialSync,
-      setMantaWalletInitialSync
+      setMantaWalletInitialSync,
+      suggestedMinFeeBalance,
+      setSuggestedMinFeeBalance
     }),
     [
       usingMantaWallet,
       setUsingMantaWallet,
       mantaWalletInitialSync,
-      setMantaWalletInitialSync
+      setMantaWalletInitialSync,
+      suggestedMinFeeBalance,
+      setSuggestedMinFeeBalance
     ]
   );
 
--- a/src/contexts/mantaWalletContext.tsx
+++ b/src/contexts/mantaWalletContext.tsx
@@ -110,7 +110,7 @@ export const MantaWalletContextProvider = ({
   useEffect(() => {
     const interval = setInterval(async () => {
       getEstimatedMinFee();
-    }, 5000);
+    }, 10000);
     return () => {
       interval && clearInterval(interval);
     };
--- a/src/contexts/mantaWalletContext.tsx
+++ b/src/contexts/mantaWalletContext.tsx
@@ -2,6 +2,7 @@ import { SubmittableExtrinsic } from '@polkadot/api/types';
 import { EventRecord, ExtrinsicStatus } from '@polkadot/types/interfaces';
 import { BN } from 'bn.js';
 import WALLET_NAME from 'constants/WalletConstants';
+import Decimal from 'decimal.js';
 import {
   MutableRefObject,
   ReactNode,
@@ -88,6 +89,27 @@ export const MantaWalletContextProvider = ({
     );
     return mantaWallet;
   }, []);
+
+  const getEstimatedMinFee = async () => {
+    if (api && publicAddress) {
+      const dummyTx = await api.tx.balances
+        .transfer(publicAddress, 123)
+        .paymentInfo(publicAddress);
+      const fee = dummyTx.partialFee.toString() / 10 ** 12; //  senderAssetType numberOfDecimals
+      const balance = Balance.fromBaseUnits(
+        AssetType.Native(config),
+        new Decimal(fee)
+      );
+      if (balance) setSuggestedMinFeeBalance(balance);
+      return balance;
+    }
+    return undefined;
+  };
+
+  // initial setSuggestedMin gas fee
+  useEffect(() => {
+    getEstimatedMinFee();
+  }, [api, publicAddress]);
 
   useEffect(() => {
     const getPrivateWallet = async () => {
@@ -204,12 +226,6 @@ export const MantaWalletContextProvider = ({
           setTxStatus(TxStatus.failed(''));
           return;
         }
-        const info = await lastTx.paymentInfo(publicAddress);
-        const [fee] = info.partialFee.toHuman().split(' ');
-        if (fee) {
-          const balance = Balance.fromBaseUnits(AssetType.Native(config), fee);
-          setSuggestedMinFeeBalance(balance);
-        }
         await lastTx.signAndSend(publicAddress, finalTxResHandler.current);
         setTxStatus(TxStatus.processing(null, lastTx.hash.toString()));
       } catch (e: any) {
@@ -224,7 +240,7 @@ export const MantaWalletContextProvider = ({
       try {
         const internalTx: any = txQueue.current.shift();
         await internalTx.signAndSend(publicAddress, handleInternalTxRes);
-      } catch (e:any) {
+      } catch (e: any) {
         setTxStatus(TxStatus.failed(`internalTx failed: ${e?.message}`));
         txQueue.current = [];
       }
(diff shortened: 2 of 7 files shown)


src/contexts/mantaWalletContext.tsx
```diff
@@ -88,6 +91,34 @@ export const MantaWalletContextProvider = ({
...
+      const fee =
+        (dummyTx.partialFee.toString() / 10 ** numberOfDecimals) * 1.1 + 0.3; // KMA
+      const balance = Balance.fromBaseUnits(
+        AssetType.Native(config),
+        new Decimal(fee)
+      );
+      if (balance) setSuggestedMinFeeBalance(balance);
+      return balance;
+    }
+    return undefined;
```
Kevingislason, line 108: It would be better to return null

src/contexts/mantaWalletContext.tsx
```diff
@@ -88,6 +91,34 @@ export const MantaWalletContextProvider = ({
     return mantaWallet;
   }, []);
 
+  const getEstimatedMinFee = async () => {
+    if (api && publicAddress) {
+      const dummyTx = await api.tx.balances
+        .transfer(publicAddress, 123)
+        .paymentInfo(publicAddress);
+      const fee =
+        (dummyTx.partialFee.toString() / 10 ** numberOfDecimals) * 1.1 + 0.3; // KMA
```
Kevingislason, line 100: I don't understand how we can do math on `dummyTx.partialFee.toString()`, since it is a string and not a number, BN, or Decimal

src/contexts/mantaWalletContext.tsx
```diff
@@ -56,12 +57,14 @@ export const MantaWalletContextProvider = ({
...
-  const { NETWORK_NAME: network } = useConfig();
-  const { usingMantaWallet } = useGlobal();
+  const config = useConfig();
+  const { NETWORK_NAME: network } = config;
+  const { usingMantaWallet, setSuggestedMinFeeBalance } = useGlobal();
   const { api } = useSubstrate();
   const { externalAccount } = usePublicAccount();
   const publicAddress = externalAccount?.address;
   const { setTxStatus } = useTxStatus();
+  const { numberOfDecimals } = AssetType.Native(config);
```
Kevingislason, line 67: I think it would be better to just define and use this variable locally inside of `getEstimatedMinFee`, since it isn't used anywhere else, and since its meaning isn't really clear outside of that context.

src/pages/SendPage/SendContext/index.tsx
```diff
@@ -646,6 +640,23 @@ export const SendContextProvider = (props) => {
...
+  }, 10000);
+
+  useEffect(() => {
+    debouncedGetEstimatedMinFee();
+  }, [
+    senderAssetCurrentBalance,
+    senderAssetTargetBalance,
+    isToPrivate(),
+    isPublicTransfer()
+  ]);
```
Kevingislason, line 658: Can you explain what `useDebouncedCallback` does here? It's a bit unusual, so you may want to include a comment in the code here.

src/pages/SendPage/SendButton.tsx
```diff
@@ -126,8 +126,8 @@ const ValidationSendButton = ({ showModal }) => {
     validationMsg = 'Connecting to network';
   } else if (!senderAssetTargetBalance) {
     validationMsg = 'Enter amount';
-  } else if (userCanPayFee() === false) {
-    validationMsg = `Insufficient ${senderNativeTokenPublicBalance?.assetType?.baseTicker} to pay transaction fee`;
```
Kevingislason, line 130: I don't think we want to get rid of this case. `userCanPayFee() === false` is different from `txWouldDepleteSuggestedMinFeeBalance()`. In the first case, the user can't afford to pay the current transaction fee. In the second case, the user can afford to pay the current transaction fee, but _possibly_ should transact for a lower amount.

src/pages/SendPage/SendButton.tsx
```diff
@@ -126,8 +126,8 @@ const ValidationSendButton = ({ showModal }) => {
...
+  } else if (txWouldDepleteSuggestedMinFeeBalance()) {
+    validationMsg = `Insufficient ${senderNativeTokenPublicBalance?.assetType?.baseTicker} for transaction fee`;
```
Kevingislason, line 130: I don't think we should handle `txWouldDepleteSuggestedMinFeeBalance` in the SendButton here, since it will prevent the user from transacting. I think we only want to show a warning in `SendErrorDisplay.tsx`. In some cases, it may make sense for the user to transact even in the transaction would delete their minimum suggested fee balance.

src/pages/SendPage/SendContext/index.tsx
```diff
@@ -391,28 +396,19 @@ export const SendContextProvider = (props) => {
   // accidentally becoming unable to transact because they cannot pay fees
   const txWouldDepleteSuggestedMinFeeBalance = () => {
     if (
-      senderAssetCurrentBalance?.assetType.isNativeToken &&
-      senderAssetTargetBalance?.assetType.isNativeToken &&
```
Kevingislason, line 395: If I recall correctly, `suggestedMinFeeBalance` was originally added to prevent users from privatizing so much Calamari that they can no longer pay fees. Removing these conditions will cause the the fee warning to display when the user is running low on fees from transacting in tokens besides Dolphin. Are you sure this change in behavior is intended?

src/contexts/mantaWalletContext.tsx
Kevingislason, line 100: Will this same formula also work for Dolphin and Manta?

src/pages/SendPage/SendContext/index.tsx
```diff
@@ -391,28 +396,19 @@ export const SendContextProvider = (props) => {
...
-        );
-      } else {
-        throw new Error('Unknown network');
-      }
-      const balanceAfterTx = senderAssetCurrentBalance.sub(
-        senderAssetTargetBalance
-      );
-      return suggestedMinFeeBalance.gte(balanceAfterTx);
+      const balanceAfterTx = senderAssetTargetBalance?.assetType
+        ? senderAssetCurrentBalance.sub(senderAssetTargetBalance)
```
Kevingislason, line 404: This function should just return `false` immediately if there is no `senderAssetTargetBalance`

src/pages/SendPage/SendContext/index.tsx
```diff
@@ -468,8 +464,6 @@ export const SendContextProvider = (props) => {
       receiverAddress &&
       senderAssetTargetBalance &&
       senderAssetCurrentBalance &&
-      userHasSufficientFunds() &&
-      userCanPayFee() &&
```
Kevingislason, line 472: Why are we removing these cases from `isValidToSend`? I don't think it is ever valid to send if either of these is false.

src/pages/SendPage/SendContext/index.tsx
```diff
@@ -646,6 +640,23 @@ export const SendContextProvider = (props) => {
     return isPublicTransfer() || isToPublic();
   };
 
+  /**
+   *
+   * Gas fee updates
+   */
+  const debouncedGetEstimatedMinFee = useDebouncedCallback(() => {
```
Kevingislason, line 647: It would be convenient to group together all of the code related to min fee balances.

src/contexts/globalContexts.tsx
```diff
@@ -19,6 +19,7 @@ const GlobalContextProvider = ({ children }) => {
   const initValue = store.get(localStorageKeys.UsingMantaWallet, true);
   const [usingMantaWallet, _setUsingMantaWallet] = useState(initValue);
   const [mantaWalletInitialSync, _setMantaWalletInitialSync] = useState(true);
+  const [suggestedMinFeeBalance, setSuggestedMinFeeBalance] = useState();
```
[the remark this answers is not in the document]
jincan39: image
yes good question it's due to the context nested level. we should put suggestedMinFeeBalance into a higher context than MantaWalletContextProvider to avoid wiped. If we want to put in PrivateWalletContext we should swap the order `PrivateWalletImplementation` & `PrivateWalletContextProvider` but it not seems ok.

src/contexts/mantaWalletContext.tsx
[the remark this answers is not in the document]
jincan39: Good advice. I'll try with Joly.

jincan39: javascript automatically convert types string to number when we do calculation. yeah in the typescript context we should notice types. `Number(dummyTx.partialFee.toString())`

src/pages/SendPage/SendContext/index.tsx
jincan39: Yeah, the gas fee will changed depends on network's status. We want to fetch the estimated gas fee immediately(but not so frequently) when user do inputing.

src/pages/SendPage/SendButton.tsx
jincan39: yeah a bit confusion here, especially in the select max amount case. I think senderAssetCurrentBalance.valueOverExistentialDeposit() has helped user to leave the min deposit ? and we maybe can get ride of the `getReservedNativeTokenBalance` .

src/contexts/mantaWalletContext.tsx
jincan39: yes think it's just numberOfDecimals difference

src/pages/SendPage/SendContext/index.tsx
jincan39: Thanks for the patience comments, good advice.

PR #1034, opened 2023-05-15
feat: add notification & subscribe api

### LINKED ISSUE #1027: MantaPay's current select network needs to be consistent with Wallet's network
If the currently selected web page is Calamari and the wallet selected is Manta or Dolphin, prompt the user to switch the wallet before allowing access.

jincan39:
## Description

- add `mantaWallet.subscribeAccounts` api
- add MantaWallet update notification
- add change network notification. close #1027 

https://www.figma.com/file/9nAYHsYlfHmyT8RL5P8ISZ/Manta-Pay?type=design&node-id=4904-2631&t=jrvPgxfehf3yxyww-0

---

## If PR to `staging`


src/contexts/globalContexts.tsx
```diff
@@ -15,10 +17,16 @@ import { KeyringContextProvider } from './keyringContext';
 
 const GlobalContext = createContext();
 
+const minMantaWalletVersion = '0.0.11';
+
 const GlobalContextProvider = ({ children }) => {
   const initValue = store.get(localStorageKeys.UsingMantaWallet, true);
   const [usingMantaWallet, _setUsingMantaWallet] = useState(initValue);
-  const [mantaWalletInitialSync, _setMantaWalletInitialSync] = useState(true);
+  const [mantaWalletVersion, setMantaWalletVersion] = useState<string | null>(
```
Kevingislason, line 25: Manta Wallet version should probably go in MantaWalletContext. In general, we want to avoid adding properties to GlobalContext unless they are absolutely necessary throughout the app. If we add a lot of unrelated data to this context for convenience, it'll become difficult to work on.

src/contexts/globalContexts.tsx
```diff
@@ -15,10 +17,16 @@ import { KeyringContextProvider } from './keyringContext';
...
+    null
+  );
+  const [shouldUpdateWallet, setShouldUpdateWallet] = useState(false);
```
Kevingislason, line 28: I would call this `mantaWalletIsOutOfDate` for clarity; we use the word wallet in quite a few senses, so it's probably best to be a bit verbose.

src/contexts/globalContexts.tsx
```diff
@@ -15,10 +17,16 @@ import { KeyringContextProvider } from './keyringContext';
...
+  const [mantaWalletInitialSync, _setMantaWalletInitialSync] = useState(false);
```
Kevingislason, line 29: This also probably belongs in `MantaWalletContext`. We already have an `isInitialSync` variable there.

src/contexts/globalContexts.tsx
```diff
@@ -15,10 +17,16 @@ import { KeyringContextProvider } from './keyringContext';
 
 const GlobalContext = createContext();
 
+const minMantaWalletVersion = '0.0.11';
```
Kevingislason, line 20: We have a variable `MIN_REQUIRED_WALLET_VERSION` in `config/common.json` for this

src/contexts/mantaWalletContext.tsx
```diff
@@ -98,6 +103,9 @@ export const MantaWalletContextProvider = ({
         if (mantaWallet?.extension?.privateWallet) {
           setPrivateWallet(mantaWallet?.extension?.privateWallet);
         }
+        const version = mantaWallet?.extension?.version;
```
Kevingislason, line 106: We have a `Version` class that you may want to use here

src/contexts/globalContexts.tsx
```diff
@@ -30,6 +38,17 @@ const GlobalContextProvider = ({ children }) => {
     _setUsingMantaWallet(initValue);
   }, [isDolphinPage]);
 
+  useEffect(() => {
+    if (mantaWalletVersion < minMantaWalletVersion) {
+      setShouldUpdateWallet(true);
```
Kevingislason, line 43: I'm not sure if comparing `mantaWalletVersion` to `minMantaWalletVersion` with `<` will work in all cases, since the versions are strings. You will probably want to use the `gte` method on the `Version` class to do this comparison.

src/contexts/globalContexts.tsx
```diff
@@ -30,6 +38,17 @@ const GlobalContextProvider = ({ children }) => {
...
+      Notification({
+        // eslint-disable-next-line react/react-in-jsx-scope
+        message: <UpdateWarningMessage />,
+        duration: 0
+      });
+    }
+  }, [mantaWalletVersion]);
```
Kevingislason, line 50: GlobalContext shouldn't be responsible for handling any UI elements. I think code this probably belongs on the `SendPage`.

src/contexts/globalContexts.tsx
jincan39: Agree, current MantaWalletContext will be render for several times which could cause multiple messages.

src/contexts/globalContexts.tsx
jincan39: Agree, currently MantaWalletContext initially will be render for several times which would be not a good place to put global variable.

src/contexts/globalContexts.tsx
jincan39: think ` versionIsOutOfDate(config.MIN_REQUIRED_WALLET_VERSION, extensionVersion)` works great.

src/contexts/globalContexts.tsx
```diff
@@ -15,6 +15,7 @@ import { KeyringContextProvider } from './keyringContext';
 
 const GlobalContext = createContext();
 
+
```
jasonintju, line 18: blank line should be removed, so this file will keep unchanged

jasonintju (commented):
render 2 times, do we have a better solution to fix this?

jincan39:
> render 2 times, do we have a better solution to fix this?

yes, we should look carefully to the context and it's dependency usage. For gradually solving this problem, i'd like to bring [zustand](https://github.com/pmndrs/zustand) to our manta land.

jasonintju (approved):

src/contexts/mantaWalletContext.tsx
```diff
@@ -104,12 +111,27 @@ export const MantaWalletContextProvider = ({
...
-      if (!privateAddress) {
-        const mantaWallet = await getMantaWallet();
-        if (!mantaWallet || !privateWallet) {
-          return;
-        }
+      const mantaWallet = await getMantaWallet();
+      if (!mantaWallet || !privateWallet || !extensionVersion) {
+        return;
+      }
+      if (extensionVersion.gte(new Version('0.0.12'))) {
```
Kevingislason, line 120: It would be better to get this version from the `config`

src/pages/SendPage/index.tsx
```diff
@@ -1,11 +1,42 @@
...
 import SendForm from './SendForm';
 import { PrivateTxHistoryContextProvider } from './privateTxHistoryContext';
 
 const SendPage = () => {
+  const config = useConfig();
+  const { extensionVersion } = usePublicAccount();
+  const outdated = versionIsOutOfDate(
+    config.MIN_REQUIRED_WALLET_VERSION,
+    extensionVersion
+  );
```
Kevingislason, line 21: I think we probably want to get and store this variable as `mantaWalletIsOutOfDate` in MantaWalletContext for convenience, since we are also using the same calculation in `src/pages/SendPage/SendButton.tsx`

src/pages/BridgePage/SendButton.tsx
```diff
@@ -48,6 +51,10 @@ const ValidationButton = () => {
     validationMsg = 'Connecting to network';
   } else if (evmIsEnabled && originChainIsEvm && chainId !== Chain.Moonriver(config).ethChainId) {
     isSwitchNetwork = true;
+  } else if (  usingMantaWallet &&
+    extensionName === WALLET_NAME.MANTA &&
+    versionIsOutOfDate(config.MIN_REQUIRED_WALLET_VERSION, extensionVersion)) {
+    validationMsg = 'Update Manta Wallet';
```
Kevingislason, line 57: Unless Chloe thinks so, I don't think we want to add this validation to the bridge page, since the wallet being out of date doesn't prevent users from bridging.

Kevingislason:
> For gradually solving this problem, i'd suggest to bring [zustand](https://github.com/pmndrs/zustand) to our manta land. #1040

On reflection, have a few concerns about this. 

First, it's going to take a lot of time to migrate all of our state management to a different library, even if it may have advantages. If we have two different state management systems throughout the app, it will make the code more difficult to understand. 

Second and more importantly, it's good in some ways to have totally separate instances of this context and certain other contexts for each different network. The app architecture does this on purpose to prevent a dangerous class of bugs where testnet and mainnet funds are mixed up.

src/contexts/mantaWalletContext.tsx
jasonintju: > It would be better to get this version from the `config`

It is not configurable. The version '0.0.12' is a CONSTANT. `gte(0.0.12)`, we will use `subscribe` api, otherwise, we will use `getAccounts` api.

src/pages/BridgePage/SendButton.tsx
jincan39: update by #1042

src/pages/SendPage/index.tsx
jincan39: Good advice, update by #1042.

jasonintju:
> > For gradually solving this problem, i'd suggest to bring [zustand](https://github.com/pmndrs/zustand) to our manta land. #1040
> 
> On reflection, have a few concerns about this.

So for now, perhaps we'd better revert zustand PR? After all, it didn't solve any issues for now.

jincan39:
> > > For gradually solving this problem, i'd suggest to bring [zustand](https://github.com/pmndrs/zustand) to our manta land. #1040
> > 
> > 
> > On reflection, have a few concerns about this.
> 
> So for now, perhaps we'd better revert zustand PR? After all, it didn't solve any issues for now.

Push a revert pr #1043

src/pages/SendPage/index.tsx
jasonintju: > I think we probably want to get and store this variable as `mantaWalletIsOutOfDate` in MantaWalletContext for convenience, since we are also using the same calculation in `src/pages/SendPage/SendButton.tsx`

I don't think so. We should only keep necessary state in context. Here, `mantaWalletVersion` is a state, but `mantaWalletIsOutOfDate` is a calculation result base on `mantaWalletVersion` state. Even we use `mantaWalletIsOutOfDate` in other places, but it is still a calculation with a util method `versionIsOutOfDate`.

src/contexts/mantaWalletContext.tsx
Kevingislason: Right, this will always be the same. We should probably assign this to a constant though, so that it's clear why this version is meaningful.

```js
const SUBSCRIBE_ACCOUNTS_FIRST_VERSION = new Version('0.0.12');
if (extensionVersion.gte(SUBSCRIBE_ACCOUNTS_FIRST_VERSION)) {
```

PR #1064, opened 2023-05-22
fix: init keyring

jasonintju:
## Description

closes: #1065 

---

## If PR to `staging`

## If PR to `main`
- [ ] Update `CHANGELOG.md`

### DIFF
--- a/src/contexts/keyringContext.tsx
+++ b/src/contexts/keyringContext.tsx
@@ -27,6 +27,7 @@ import {
 } from 'utils/persistence/walletStorage';
 import isObjectEmpty from 'utils/validation/isEmpty';
 import { useActive } from 'hooks/useActive';
+import { cryptoWaitReady } from '@polkadot/util-crypto';
 
 type KeyringContextValue = {
   keyring: Keyring;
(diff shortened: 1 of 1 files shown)


Kevingislason (commented):
Instead of replacing the `web3ExtensionInjected` logic with cryptoWaitReady, I think we may want to keep the old `web3ExtensionInjected` logic and also add the `cryptoWaitReady` check.

The point of `web3ExtensionInjected`is to check that the wallet extensions are available in browser, which takes an unpredictable amount of time. `cryptoWaitReady` checks that polkadot.js WASM libraries are ready. We want to be sure that both of unrelated conditions are met. As far as we know, `cryptoWaitReady`  is sometimes faster than the `web3ExtensionInjected` check and sometimes slower, so imo we should keep both checks to be safe.

jasonintju:
> Instead of replacing the `web3ExtensionInjected` logic with cryptoWaitReady, I think we may want to keep the old `web3ExtensionInjected` logic and also add the `cryptoWaitReady` check.
> 
> The point of `web3ExtensionInjected`is to check that the wallet extensions are available in browser, which takes an unpredictable amount of time. `cryptoWaitReady` checks that polkadot.js WASM libraries are ready. We want to be sure that both of unrelated conditions are met. As far as we know, `cryptoWaitReady` is sometimes faster than the `web3ExtensionInjected` check and sometimes slower, so imo we should keep both checks to be safe.

I'm not sure about why to check `web3ExtensionInjected`. Even there is no extensions, we still can initialize the keyring, right?

PR #1112, opened 2023-06-15
estimate private tx bacthes

Kevingislason:
## Description

Estimates number of transaction batches required when estimating fees

---

## If PR to `staging`


src/pages/SendPage/SendContext/index.tsx
```diff
@@ -647,67 +650,137 @@ export const SendContextProvider = (props) => {
...
+      } else if (isPrivateTransfer()) {
+        tx = api.tx(dummyPrivateTransfer);
+      } else if (isToPublic()) {
+        tx = api.tx(dummyToPublic);
+      }
+      return tx;
+    };
+
+    const getFeeEstimate = async () => {
+      console.log('getFeeEstimate');
```
jasonintju, line 747: a console here

jasonintju (dismissed):

src/contexts/mantaWalletContext.tsx
```diff
@@ -435,6 +433,39 @@ export const MantaWalletContextProvider = ({
...
+        const transferPostsCount = await privateWallet?.estimateTransferPostsCount(
+          {
+            network,
+            transactionType,
+            assetId: targetBalance.assetType.assetId.toString(),
+            amount: targetBalance.valueAtomicUnits.toString(),
+            zkAddressOrPolkadotAddress: address,
+          }
+        );
+        return Math.ceil(transferPostsCount / 2);
```
jasonintju, line 462: Here, I think the gas fee you mocked includes `two transferPosts`. So how to mock the gas fee that only includes `one transferPost`?

jasonintju:
Btw, the `transferPostsCount` returned from the api now are wrong. We have to wait for runtime check.

src/contexts/mantaWalletContext.tsx
Kevingislason: I think all transactions contain two transfer posts, but I'm not 100% sure

src/constants/DummyTransactions.ts
```diff
@@ -1,3 +1,8 @@
+
```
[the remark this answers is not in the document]
Kevingislason: I added a log for each transaction type to log a serialized transaction. Then I did some arbitrary transactions locally, looked at the console, and copy-pasted those serialized transactions. The size should be the same for all transactions without regard to parameters, unless the number of transfer posts per batch are variable ( I don't think they are).

jasonintju (approved):

